PR #1, opened 2017-05-12
Make it easy to verify the shell scripts with shellcheck

bvanassche:


osandov:
I like this, can we make the default Makefile target not be shellcheck, though? For now, we can just have a dummy `all` target that does nothing.

bvanassche:
Done :-) Can you have a look at the updated merge request?

PR #4, opened 2017-10-10
Nishita

nishita28:
This pull request contains the block integrity patches. Tests 14 and 15 panic the kernel for versions 4.9 and above. On doing a git bisect, it pointed me to "25f4c41415e513f0e9fb1f3fce2ce98fcba8d263 (block: implement (some of)  fallocate for block devices)"as the offending commit. I am working on getting the root cause.

Thanks,
Nishita


tests/block/009
```diff
@@ -0,0 +1,54 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+DESCRIPTION="check page-cache coherency after BLKDISCARD"
+
+. ./common/rc
+. ./common/scsi_debug
```
osandov: Really minor nit, but just make this
```
. common/rc
. common/scsi_debug
```
for consistency with the other tests.

tests/block/009
```diff
@@ -0,0 +1,54 @@
...
+
+requires() {
+	_have_scsi_debug && _have_program blkdiscard && _have_program xfs_io
```
osandov: blkdiscard comes from util-linux which is required to be installed, so no need to have an explicit check for it.

tests/block/009
```diff
@@ -0,0 +1,54 @@
...
+}
+
+test_device() {
```
osandov: This test creates its own device instead of using `$TEST_DEV`, so this should be `test()` and not `test_device()`.

common/scsi_debug
```diff
@@ -65,3 +65,30 @@ _exit_scsi_debug() {
 	unset SCSI_DEBUG_DEVICES
 	modprobe -r scsi_debug
 }
+
+# Args: [physical sector size, [logical sector size, [unaligned(0|1), [size in megs]]]]
+_get_scsi_debug_dev()
```
osandov: There's already an `_init_scsi_debug` helper, please use that instead of adding this one.

tests/block/009
```diff
@@ -0,0 +1,54 @@
...
+	echo "Running ${TEST_NAME}"
+
+	rm -f $seqres.full
+
+	# Create virtual device with unmap_zeroes_data support
+	dev=$(_get_scsi_debug_dev 512 512 0 32 "lbpws=1 lbpws10=1")
+
+	# Initialize data with known pattern
+	xfs_io -c "pwrite -S 0xaa -b 2M 0 32M -w" -d $dev >>$seqres.full 2>&1 || \
```
osandov: `$seqres.full` is called `$FULL` in blktests. Additionally, get rid of the `2>&1 || _fail` here and below. If xfs_io fails, it will write to stderr and will fail anyways since it doesn't match the golden output.

common/liotarget
```diff
@@ -0,0 +1,110 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+
+#
+# Functions for Linux-IO Target manipulation
+#
+
+_require_liotarget()
```
osandov: This should be `_have_liotarget`, and it should work like the other `_have_*` examples in common/rc. Namely, `_notrun` won't work for blktests, you need to return 0 if the requirements are met or set `SKIP_REASON` and return 1 if not.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
 	readlink -f "$TEST_DEV_SYSFS/device" | \
 		grep -Eo '[0-9a-f]{4}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]'
 }
+
+_notrun()
```
osandov: Like I noted above, this won't work in blktests. The `_have_*` helpers are how you do this in blktests.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+	echo "$*" > $seqres.notrun
+	echo "$seq not run: $*"
+	rm -f ${RESULT_DIR}/require_test*
+	rm -f ${RESULT_DIR}/require_scratch*
+
+	status=0
+	exit
+}
+
+_fail()
```
osandov: This also won't work in blktests. If helpers are calling it, they should just print an error and return 1 instead, and tests usually shouldn't be exiting.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+{
+	echo "$*" | tee -a $seqres.full
+	echo "(see $seqres.full for details)"
+	status=1
+	exit 1
+}
+
+_require_loop()
```
osandov: We already have `_have_loop`.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+	modprobe loop >/dev/null 2>&1
+	if grep loop /proc/devices >/dev/null 2>&1
+	then
+		:
+	else
+		_notrun "This test requires loopback device support"
+	fi
+}
+
+_create_loop_device()
```
osandov: This helper isn't very helpful, just use losetup directly.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+{
+	file=$1
+	dev=`losetup -f --show $file` || _fail "Cannot assign $file to a loop device"
+	echo $dev
+}
+
+_destroy_loop_device()
```
osandov: Same, just use losetup directly. See the tests in tests/loop for examples.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+{
+	dev=$1
+	losetup -D $dev || _fail "Cannot destroy loop device $dev"
+}
+
+_mkfs_dev()
```
osandov: This is way too much xfstests baggage. If a test really needs this, then it probably belongs in xfstests and not blktests. Otherwise just call mkfs directly.

common/rc
```diff
@@ -140,3 +143,101 @@ _get_pci_dev_from_blkdev() {
...
+		# output stored mkfs output
+		cat $tmp_dir.mkfserr >&2
+		cat $tmp_dir.mkfsstd
+		status=1
+		exit 1
+	fi
+	rm -f $tmp_dir.mkfserr $tmp_dir.mkfsstd
+}
+
+_mount()
```
osandov: Just use mount directly.

common/scsi_debug
nishita28: _get_scsi_debug_dev takes some options as input and returns a device back. Should _init_scsi_debug be modified to do something similar?

nishita28:
Thanks for your feedback. 

1. Should block/014 and block/015 be pushed to xfstests then?

2. I am working on making block/012 and block/013 more like the existing tests, will get in touch incase I need help

common/scsi_debug
osandov: You can pass arbitrary module parameters to `_init_scsi_debug` and it'll pass them along to modprobe. `_init_scsi_debug` sets `SCSI_DEBUG_DEVICES` instead of outputting a single block device. So for your test you can do
```
_init_scsi_debug lbpws=1 lbpws10=1
local dev="/dev/${SCSI_DEBUG_DEVICES[0]}"
```

osandov:
Yeah, block/014 and block/015 make more sense as a generic xfstest if anything.

nishita28:
I have updated the pull request with the necessary changes to test 009.

Thanks,
Nishita

nishita28:
Modified test 009 as per your suggestion. I have made changes to tests 012 and 013 to make them more like blktests. Once you get 009 in, I shall send a pull request for those 2 as well. 

Thanks,
Nishita

osandov:
Thanks, this looks good. One last thing: could you rebase this onto my current master branch?
 It still has the fsstress commit on there which is not needed. I also need your Signed-off-by: in the commit.

nishita28:
I submitted a new pull request with changes rebased against master https://github.com/osandov/blktests/pull/6

PR #6, opened 2017-11-02
add: Check page-cache coherency after BLKDISCARD

nishita28:
Regression test for fix https://lkml.org/lkml/2017/3/22/789
Page cache should be dropped after successful discard

Signed-off-by: Dmitry Monakhov <dmonakhov@openvz.org>
Signed-off-by: Nishita Borkar <nishita.borkar@oracle.com>


osandov:
Thanks, Nishita!

PR #7, opened 2017-11-08
add test for read-only state

ddiss:


tests/block/009
```diff
@@ -25,7 +25,7 @@ DESCRIPTION="check page-cache coherency after BLKDISCARD"
 . common/scsi_debug
 
 requires() {
-	_have_scsi_debug && _have_program xfs_io
+	_have_scsi_debug && _have_program xfs_io && _have_program blkdiscard
```
osandov: blkdiscard is part of util-linux, which we already expect to be installed. Or are there versions of util-linux without blkdiscard floating around?

src/sg/dxfer-from-dev.c
```diff
@@ -51,7 +51,8 @@ int main(int argc, char **argv)
 	}
 
 	printf("PASS\n");
-
+	rc = 0;
 out_close:
 	close(fd);
+	return (rc ? 1 : 0);
```
osandov: Weird, I don't get the warning, but I'll apply this.

tests/block/012
```diff
@@ -0,0 +1,45 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+DESCRIPTION="check read-only state"
+TIMED=1
+
+requires() {
+	_have_program xfs_io && _have_program blockdev
```
osandov: Same as the note for blkdiscard, blockdev is part of util-linux.

tests/block/012
```diff
@@ -0,0 +1,45 @@
...
+}
+
+test_device() {
+	echo "Running ${TEST_NAME}"
+
+	blockdev --getro "$TEST_DEV"
+	xfs_io -c "pwrite -w -S 0xaa -b 2M 0 2M" -d "$TEST_DEV" >>"$FULL"
+	dd if="$TEST_DEV" bs=2M count=1 2>>"$FULL" | hexdump
+	blockdev --setro "$TEST_DEV"
+	blockdev --setro "$TEST_DEV"
```
osandov: Any particular reason you do --setro twice here? Looks good otherwise, thanks!

tests/block/009
ddiss: thanks for the feedback, Omar. I run blktests from an initramfs image ( https://github.com/rapido-linux/rapido/blob/master/cut_blktests_zram.sh and cut_blktests_rbd.sh ), so this was just intended as a reminder. I'll drop it along with the blockdev one below.

tests/block/012
ddiss: the intention here was just to confirm that BLKROSET succeeds while the device is already flagged read-only.

ddiss:
Any feedback on the new version? I dropped the unnecessary _have_program() calls, but kept the 2x --setro.

osandov:
Merged, thanks! I had to add an extra change to deal with my newer version of xfs_io using pwrite instead of pwrite64, could you double check that it still passes for you?

ddiss:
Thanks for fixing that... I'm running with xfsprogs 4.3.0

PR #8, opened 2017-11-10
liotarget block tests 012 and 013

nishita28:
I have modified the lio tests to make them work more like existing blktests. Let me know your comments.

Thanks,
Nishita


osandov:
Do you have documentation for how to get these tests running? This is as far as I got:

```
# targetcli --help
The target service is not running.
```

nishita28:
I did not really follow any documentation for this. Just installed targetcli by doing "yum install targetcli" and the tests run fine. Which Linux variant are you seeing this on? Also, can you tell me which version of targetcli is running?

ddiss:
> Do you have documentation for how to get these tests running? This is as far as I got:
> 
> # targetcli --help
> The target service is not running.

You need to have the target_core_mod and configfs kernel modules loaded. FWIW, it looks as though you're using the old Datera targetcli package, rather than the targetcli-fb fork.

I'd be glad to see LIO tests added to blktests, but would be in favour of having blktests manually configure LIO via the raw configfs paths, rather than pulling in new targetcli[-fb] + rtslib[-fb] dependencies.

osandov:
Thanks for the tip, @ddiss, the -fb fork at least runs and also happens to be packaged for my distro. However, I agree that we should just do the setup with configfs directly. @nishita28, would you be able to do that?

nishita28:
@osandov, sure I will do that. I shall ping you if I have any questions.

ddiss:
@nishita28 do you need any help with your configfs rework here? I'm looking forward to having LIO coverage included in blktests.

bvanassche:
@nishita28, in addition to the help offered by @ddiss: examples of how to configure LIO through configfs are available in https://github.com/osandov/blktests/pull/24.

osandov:
This hasn't seen activity in awhile. Feel free to reopen if you get back to this.

PR #9, opened 2017-11-22
block/006, block/008, block/010: add _have_fio to require function

yide1728:
Add the missed _have_fio for block/006, block/008, block/010

Signed-off-by: Yi Zhang <yi.zhang@redhat.com>


osandov:
Merged, thanks, @yide1728 and @ddiss!

PR #10, opened 2018-01-02
Append not replace cpus to cpus hotplug list

felixsch:
Due to the missing plus, only the last cpu in directory listing was used in hotpluggable cpus.


osandov:
Nice catch, thanks. Could you please add your Signed-off-by as per https://github.com/osandov/blktests/blob/master/CONTRIBUTING.md. The description in your pull request should also go in the commit message, as well.

felixsch:
@osandov Sorry for the long delay. Added the signoff!

osandov:
Thanks for the fix!

PR #12, opened 2018-03-01
dir-locals.el: add emacs local variable config

morbidrsa:
Add a dir-locals.el so Emacs will do the right thing when editing
blktests code and test cases.


osandov:
Thanks!

PR #13, opened 2018-03-07
sg: test for patch "Keep disk read-only when re-reading partition"

jeremycline:
Add a regression test for the patch "scsi: sd: Keep disk read-only when
re-reading partition"[0] which ensures re-reading a disk's partition table
doesn't change its read-only flag.

[0] https://www.spinics.net/lists/linux-scsi/msg118162.html


osandov:
Looks great, although it's an sd test rather than sg. I'll merge this and just rename the category to scsi.

PR #14, opened 2018-03-23
Test big discovery log pages

morbidrsa:
This pull adds a new test-case to the nvme test group for testig big discovery log pages.

It is inspired by a bug triggered in the user-space 'nvme discover' with a non Linux target which set's an non 0 MDTS (Maximum Data Transfer Size) value causing nvme cli to fail the discovery.

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -24,3 +24,68 @@ _test_dev_is_nvme() {
 	fi
 	return 0
 }
+
+_create_nvmet_port()
+{
+	trtype=$1
+	nvmet_cfs="/sys/kernel/config/nvmet/"
+
+	port=$(ls ${nvmet_cfs} | wc -l)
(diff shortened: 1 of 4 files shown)


osandov:
Is there a userspace commit to reference from the test?

morbidrsa:
Not really yet and you wouldn't break my heart if you'd agree to xchg test/nvme/002 from this pull with the test-case from here: http://lists.infradead.org/pipermail/linux-nvme/2018-March/016466.html I have no problem updating the pull request.

OTOH it's still a nice scalability test and we didn't do enough of these  to NVMe upstream.

Let me know what you think.

morbidrsa:
@osandov any descission?

osandov:
Merged with some cleanups, sorry for the delay!

PR #15, opened 2018-04-09
nvme: don't require nvme devices for the whole group

morbidrsa:
Currently the nvme group requires nvme devices, but only one test
currently targets nvme devices the other tests create the devices
themselves using the loopback target.

Remove the group_device_requires() _test_dev_is_nvme to a simple
device_requires() in test/nvme/001.

### DIFF
--- a/tests/nvme/001
+++ b/tests/nvme/001
@@ -26,6 +26,10 @@ requires() {
 	_have_tracefs && _have_tracepoint "nvme/nvme_setup_nvm_cmd"
 }
 
+device_requires() {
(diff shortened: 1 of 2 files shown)


osandov:
nvme/002 and nvme/003 will still run even if you don't have an nvme device. group_device_requires() only applies to test_device() tests, and it's a nice shortcut for future tests. Is that working as expected for you or is there a problem?

morbidrsa:
Oh actually I didn't know this. Then all's OK. I'll close this PR.

Issue #17, opened 2018-04-23
block/009 failed


gurugio:
Hi,

The block/009 test case failed as below.
I tested it on two machines and both resulted the same failure.
* I tried with xfsprogs 4.9.0 on Ubuntu and xfsprogs 4.16.0-rc1 on Debian but failed
* I tried "noop" and "deadline" IO scheduler but failed

```
gohkim@ws00837:~/work/tools/blktests$ sudo ./check block/009
block/009 (check page-cache coherency after BLKDISCARD)      [failed]
    runtime  ...  0,977s
    --- tests/block/009.out	2018-04-23 17:52:00.222822301 +0200
    +++ results/nodev/block/009.out.bad	2018-04-23 17:54:34.571254091 +0200
    @@ -1,8 +1,17 @@
     Running block/009
     0000000 0000 0000 0000 0000 0000 0000 0000 0000
     *
    +0200000 bbbb bbbb bbbb bbbb bbbb bbbb bbbb bbbb
    +*
    +0c00000 0000 0000 0000 0000 0000 0000 0000 0000
    +*
    ...
    (Run 'diff -u tests/block/009.out results/nodev/block/009.out.bad' to see the entire diff)
gohkim@ws00837:~/work/tools/blktests$ cat results/nodev/block/009.out.bad 
Running block/009
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0200000 bbbb bbbb bbbb bbbb bbbb bbbb bbbb bbbb
*
0c00000 0000 0000 0000 0000 0000 0000 0000 0000
*
2000000
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0200000 bbbb bbbb bbbb bbbb bbbb bbbb bbbb bbbb
*
0c00000 0000 0000 0000 0000 0000 0000 0000 0000
*
2000000
modprobe: FATAL: Module scsi_debug is in use.
Test complete
```

If I ran blkdiscard after the test, the sdg device was cleared.
```
root@ib1:/home/gkim/pb-ltp/blktests# ./check block/009
block/009 (check page-cache coherency after BLKDISCARD)      [failed]
    runtime  2.334s  ...  2.380s
    --- tests/block/009.out	2018-04-23 14:21:33.011890789 +0200
    +++ results/nodev/block/009.out.bad	2018-04-23 18:07:12.451914680 +0200
    @@ -1,8 +1,17 @@
     Running block/009
[...]

PR #18, opened 2018-05-16
Add partition table handling test

josefbacik:
Adds some more infrastructure for making sure we don't race with connect/disconnect with NBD for things that require appropriate timing, and then add a test to make sure we handle creating and deleting the partition tables properly.  This will fail without the kernel patches that I'm going to send upstream, so keep that in mind.

### DIFF
--- a/common/nbd
+++ b/common/nbd
@@ -44,5 +44,6 @@ EOF
 
 _stop_nbd_server() {
 	kill -SIGTERM "$(cat "${TMPDIR}/nbd.pid")"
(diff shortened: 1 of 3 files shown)


sitsofe:
(Is there anyway to mention the commit that nbd/002 requires in the commit message or the test? According to https://github.com/osandov/blktests/issues/17#issuecomment-384341736 sometimes there's an attempt to list the commit id in case people go using the test suite on older kernels...)

osandov:
Thanks, applied with some fixes/cleanups.

PR #23, opened 2018-05-31
common/cpuhotplug: fix the match of cpu dirs

xilabao:
If enable CONFIG_BOOTPARAM_HOTPLUG_CPU0 in kernel,
it will make CPU0 be hot pluggable, block/008 might
be failed when all CPUs are hot pluggable.

Signed-off-by: Chen Rong <chenr.fnst@cn.fujitsu.com>


osandov:
Hey, @xilabao, thanks for the report. I took a different approach for the fix, since I wanted to also exercise the case of offlining CPU 0. Let me know if it doesn't work for you.

PR #24, opened 2018-06-19
Add SRP initiator driver tests

bvanassche:


osandov:
Merged most of these, replaced some of them with a different approach. Thanks, Bart!

osandov:
Ah, actually, just noticed that you added the SRP stuff here, reopening.

PR #27, opened 2018-07-04
settle udev before unloading null_blk

gurugio:
Hi,

I am not sure it happens only on my machines but I often have "modprobe: FATAL: Module null_blk is in use." error.
So I adds "udevadm settle" before unloading null_blk module.

Second and third pathes create and apply common helpers for using null_blk module.
They do not fix anything.


osandov:
Thank you!

Issue #26, opened 2018-06-26
Some tests fail with "'device' is not a PCI device"


mrnuke:
I was trying to run the batch of tests, and noticed that some tests fail with `device is not a PCI device':

	[root@g-prime blktests]# cat config 
	TEST_DEVS=(/dev/nvme2n1)
	[root@g-prime blktests]# ./check  -d
	cat: /sys/block/nvme2c33n1/dev: No such file or directory
	[snip]
	block/011 => nvme2n1 (disable PCI device while doing I/O)    [not run]
	    /dev/nvme2n1 is not a PCI device

Looking a bit, I tried to find the logic that decides whether the device is a PCI device or not:

	[root@g-prime blktests]# source check
	^C
	[root@g-prime blktests]# readlink -f $(_find_sysfs_dir /dev/nvme2n1)
	cat: /sys/block/nvme2c33n1/dev: No such file or directory
	/sys/devices/virtual/nvme-subsystem/nvme-subsys2/nvme2n1
	[root@g-prime blktests]# readlink -f $(_find_sysfs_dir /dev/nvme2)
	/sys/devices/pci0000:ae/0000:ae:00.0/0000:af:00.0/0000:b0:04.0/0000:b1:00.0/nvme/nvme2/nvme1n1

Okay, the `/dev/nvme2` node eventually traces back to a PCI device. Maybe I need to use this one, but this is also not accepted by the framework:

	[root@g-prime blktests]# cat config 
	TEST_DEVS=(/dev/nvme2)
	[root@g-prime blktests]# ./check  -d
	./check: /dev/nvme2 is not a block device

I suspect I'm either using this completely wrong, or something else is broken along the way.

osandov:
@morbidrsa can you please take a look.

morbidrsa:
@mrnuke can you give me the output of  ```tree /sys/devices/virtual/nvme-subsystem/nvme-subsys2/nvme2n1```

mrnuke:
```
#  tree /sys/devices/virtual/nvme-subsystem/nvme-subsys2/nvme2n1
/sys/devices/virtual/nvme-subsystem/nvme-subsys2/nvme2n1
├── alignment_offset
├── bdi -> ../../../bdi/259:5
├── capability
├── dev
├── device -> ../../nvme-subsys2
├── discard_alignment
├── eui
├── ext_range
├── hidden
├── holders
├── inflight
├── integrity
│   ├── device_is_integrity_capable
│   ├── format
[...]

PR #30, opened 2018-07-16
common/rc: search harder for pci device

morbidrsa:
Some multipathed NVMe devices have a bit of a different sysfs layout
then we've expoected them to have in the first place.

Try harder searching for a pci device if we encounter such a device.

This fixes issue #26.


osandov:
Fixed this:
```
common/rc:173:19: note: Double quote to prevent globbing and word splitting. [SC2086]
common/rc:174:20: note: Double quote to prevent globbing and word splitting. [SC2086]
```
And applied. Thanks!

PR #34, opened 2018-10-10
Add filter function for nvme discover

frankenmichl:
Several NVMe tests (002, 016, 017) used a pipe to a sed call filtering
the output. This call is moved to a new filter function nvme/rc and
the calls to sed are replaced by this function.

Additionally, the test nvme/016 failed for me due to the Generation
counter being greater than 1, so the new filter function was expanded to
replace the Generation counter with 'X'.

Signed-off-by: Michael Moese <mmoese@suse.de>

### DIFF
--- a/tests/nvme/017
+++ b/tests/nvme/017
@@ -38,7 +38,7 @@ test() {
 	port="$(_create_nvmet_port "loop")"
 	_add_nvmet_subsys_to_port "${port}" "${subsys_name}"
 
-	nvme discover -t loop | sed -r -e "s/portid:  [0-9]+/portid:  X/"
+	nvme discover -t loop | _filter_discovery
 	_remove_nvmet_subsystem_from_port "${port}" "${subsys_name}"
 	_remove_nvmet_port "${port}"
 
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -118,3 +118,8 @@ _find_nvme_loop_dev() {
 		fi
 	done
 }
+
+_filter_discovery() {
+	sed -r  -e "s/portid:  [0-9]+/portid:  X/" \
+		-e "s/Generation counter [0-9]+/Generation counter X/"
+}
--- a/tests/nvme/002
+++ b/tests/nvme/002
@@ -33,7 +33,7 @@ test() {
 		_add_nvmet_subsys_to_port "${port}" "blktests-subsystem-$i"
 	done
 
-	nvme discover -t loop | sed -r -e "s/portid:  [0-9]+/portid:  X/"
+	nvme discover -t loop | _filter_discovery
 
 	for ((i = iterations - 1; i >= 0; i--)); do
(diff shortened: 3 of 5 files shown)


tests/nvme/rc
```diff
@@ -118,3 +118,7 @@ _find_nvme_loop_dev() {
 		fi
 	done
 }
+
+_filter_discovery() {
```
morbidrsa: Why not:
sed -r -e "s/portid:  [0-9]+/portid:  X/" \
          -e "s/Generation counter [0..9]+/Generation counter X/"

osandov:
Regarding the generation counter issue, when I encountered it, @ChaitanayaKulkarni said that it's an actual failure, so we shouldn't filter it. Chaitanya, can you take a look?

morbidrsa:
I really wonder why, here's the definition in Fabrics 1.0:

> Generation Counter (GENCTR): Indicates the version of the discovery information,
> starting at a value of 0h. For each change in the Discovery Log, this counter is
> incremented by one. If the maximum count is exceeded, it wraps to a value of 0h.

I read it as, every time we change something that will be reflected in the discovery log page, the counter get's incremented. So I don't really see a problem in filtering it, @ChaitanayaKulkarni can you please clarify it?

osandov:
@frankenmichl do you still see the nvme/016 failure? I wasn't able to reproduce it again.

ChaitanayaKulkarni:
I'm extremely sorry guys. I wasn't aware of this conversation and I only monitor the linux-block mailing list.

Let me take a look at the testacses on latest kernel and make the fixes.

lsgunth:
Hey @ChaitanayaKulkarni , 

This still seems to be an issue even after a few months. I've found people complaining about it on the list (@tytso )  but I'm still seeing these failures in test nvme/002,016 and 017 with no obvious solution. The change in this PR seems like the only sensible solution I've heard so far.

The generation counter is a static variable in the NVMe Target code that gets incremented any time the discovery information changes. My best guess is that the kernel started incrementing this for more reasons since the tests were written and now the tests are broken. 

IMO masking these changes out as suggested in this PR is the best solution. Then, if we want to test the generation counter, create another test that ensures the counter changes after performing specific actions.

If we do that, then we can also remove the "modprobe -r" calls (which are necessary to reset the counter) and be able to support running the tests with the modules built-in (which would be very nice for people trying to test the kernel in VMs).

One way or another, we need a solution that's less fragile than the current method.

Thanks,

Logan

tests/nvme/017
```diff
@@ -38,7 +38,7 @@ test() {
 	port="$(_create_nvmet_port "loop")"
 	_add_nvmet_subsys_to_port "${port}" "${subsys_name}"
 
-	nvme discover -t loop | sed -r -e "s/portid:  [0-9]+/portid:  X/"
+	nvme discover -t loop | _filter_discovery
```
lsgunth: This change missed the corresponding update to `017.out`

osandov:
This should be fixed by the recent patches from @lsgunth.

PR #36, opened 2018-10-29
src/discontiguous-io: Do not shadow variables

bvanassche:
Avoid using variables in an inner scope with the same name as variables in
an outer scope. Enable the -Wshadow compiler flag for C and C++ source
files such that in the future the compiler will complain about shadowing.


bvanassche:
Reopening.

PR #37, opened 2018-11-01
Fix 32-bit build and add a Travis configuration file

bvanassche:


bvanassche:
For sample Travis output, see also https://travis-ci.org/bvanassche/blktests/builds/449474105.

src/Makefile
```diff
@@ -12,8 +12,9 @@ CXX_TARGETS := \
 
 TARGETS := $(C_TARGETS) $(CXX_TARGETS)
 
-CFLAGS := -O2 -Wall -Wshadow
-CXXFLAGS := -O2 -std=c++11 -Wall -Wextra -Wshadow -Wno-sign-compare -Werror
+override CFLAGS   := -O2 -Wall -Wshadow $(CFLAGS)
+override CXXFLAGS := -O2 -std=c++11 -Wall -Wextra -Wshadow -Wno-sign-compare \
+		     -Werror $(CXXFLAGS)
```
osandov: According to the [GNU make documentation](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html), this means that the flags passed on the command line won't be used at all. Should this be `+=` instead of `:=`?

osandov: You're right, I missed the `$(CFLAGS)` at the end.

PR #38, opened 2018-11-22
Fix permissions of golden output files

frankenmichl:
Four of the golden outputs had execute bits set. However, like all the
other .out-files, their permission should be restricted to 644.

Signed-off-by: Michael Moese <mmoese@suse.de>


morbidrsa:
Looks good,
Reviewed-by: Johannes Thumshirn <jthumshirn@suse.de>

osandov:
Thanks,commited. I also added a check to `make check` to prevent this in the future.

PR #39, opened 2018-12-06
Two improvements for the srp tests

bvanassche:


osandov:
Personally, I'd prefer to delete the legacy dm tests. The other change looks fine.

bvanassche:
My own preference is to do that only after the legacy block layer has been removed from all Linux distributions. The legacy block layer tests are still useful for Linux distributions that ship with a kernel before v4.21. Since kernel v4.21 has not yet been released that is the case for all Linux distributions.

common/multipath-over-rdma
```diff
@@ -14,6 +14,14 @@ if [ $ramdisk_size -gt $max_ramdisk_size ]; then
 	ramdisk_size=$max_ramdisk_size
 fi
 
+have_legacy_dm() {
```
osandov: New helpers should have a leading underscore. I.e., `_have_legacy_dm`.

common/multipath-over-rdma
```diff
@@ -14,6 +14,14 @@ if [ $ramdisk_size -gt $max_ramdisk_size ]; then
...
+	if ! _have_kernel_option DM_MQ_DEFAULT; then
+		SKIP_REASON="legacy device mapper support is missing"
+		echo "$SKIP_REASON" >/dev/null # For shellcheck
```
osandov: Add `. common/shellcheck` at the top of this file and this won't be necessary.

tests/nvmeof-mp/004
```diff
@@ -7,6 +7,10 @@
 DESCRIPTION="File I/O on top of multipath concurrently with logout and login (sq-on-mq)"
 TIMED=1
 
+requires() {
+	have_legacy_dm || return $?
```
osandov: The `|| return $?` isn't necessary here and everywhere else that it's the last command in a function.

PR #40, opened 2018-12-07
scsi/006: cache_type change may be not supported by device

im-0:
There are at least two SCSI/SAS controllers that do not support
changing of cache_type:

	HP Smart Array P410i (103c:323a, hpsa)
	Dell PERC H200 (1000:0072, mpt3sas)

When trying to change cache_type, write always fails:

	# echo "none" >cache_type
	bash: echo: write error: Invalid argument

And following appears in dmesg:

	[13007.865745] sd 1:0:1:0: [sda] Sense Key : Illegal Request [current]
	[13007.865753] sd 1:0:1:0: [sda] Add. Sense: Invalid field in parameter list

This happens even when trying to write value that is equal to current.

### DIFF
--- a/tests/scsi/006
+++ b/tests/scsi/006
@@ -27,13 +27,31 @@ test_device() {
 	)
 	local cache_type
 	local original_cache_type
+	local not_supported
+	local new_cache_type
 
 	original_cache_type="$(cat "${TEST_DEV_SYSFS}"/device/scsi_disk/*/cache_type)"
 	for cache_type in "${cache_types[@]}"; do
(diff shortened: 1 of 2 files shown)


osandov:
Coincidentally, I recently hit a similar issue with virtio-scsi. See the fix I just pushed, I think it should work for you, too.

im-0:
@osandov, your fix works too, thanks!

By the way, it looks like in at least one of my cases this was triggered by a combination of small kernel bug and too picky hardware: https://lkml.org/lkml/2018/12/23/13.

Issue #28, opened 2018-07-06
block/008 fails with "+_fio_perf: too many terse lines"


mrnuke:
Not sure if PEBKAC or incompatibility with linux-4.18-rc3

I'm having a bit of difficulty following where the "clock setaffinity failed" message originates.

```
block/008 => nvme2n1 (do IO while hotplugging CPUs)          [failed]
    runtime  39.688s  ...  38.372s
    --- tests/block/008.out     2018-07-06 14:58:35.013250320 -0500
    +++ results/nvme2n1/block/008.out.bad       2018-07-06 15:14:22.495426191 -0500
    @@ -1,2 +1,4 @@
     Running block/008
    +clock setaffinity failed: Invalid argument
    +_fio_perf: too many terse lines
     Test complete
```

The contents of "$TMPDIR/fio_perf" are:
```
clock setaffinity failed: Invalid argument
[...]

PR #41, opened 2018-12-09
Better fix for osandov/blktests#28

im-0:
I'm constantly hitting this problem:

	block/008 => sdb (do IO while hotplugging CPUs)              [failed]
	    runtime    ...  258.910s
	    --- tests/block/008.out     2018-12-06 07:52:28.971820224 +0500
	    +++ /root/blktests/results/sdb/block/008.out.bad    2018-12-09 19:25:08.973348308 +0500
	    @@ -1,2 +1,4 @@
	     Running block/008
	    +clock setaffinity failed: Invalid argument
	    +_fio_perf: too many terse lines
	     Test complete

It seems that 0.2 second sleep is not enough because my system is too slow.

Instead of adding more sleep, this patch implements proper (I hope so)
logic that waits until fio opens the test device.

Signed-off-by: Ivan Mironov <mironov.ivan@gmail.com>

### DIFF
--- a/common/fio
+++ b/common/fio
@@ -156,7 +156,7 @@ _run_fio() {
 		args+=("--runtime=$TIMEOUT")
 	fi
 
-	if ! fio "${args[@]}" "$@"; then
+	if ! BLKTESTS_MARKER=1 fio "${args[@]}" "$@"; then
 		echo "fio exited with status $?"
 		cat "$TMPDIR"/fio_perf
 	fi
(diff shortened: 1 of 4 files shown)


osandov:
Can you please try the following simpler fix?

```patch
diff --git a/tests/block/008 b/tests/block/008
index 2586e21..108f0ea 100755
--- a/tests/block/008
+++ b/tests/block/008
@@ -23,8 +23,10 @@ test_device() {
 		size="1g"
 	fi
 
-	# start fio job
-	_run_fio_rand_io --filename="$TEST_DEV" --size="$size" &
+	# Start fio job. When using the default cpu clocksource, fio does some
+	# sched_setaffinity() calls, which may race with our hotplugging. Use
+	# the clock_gettime clocksource to avoid that.
+	_run_fio_rand_io --filename="$TEST_DEV" --size="$size" --clocksource=clock_gettime &
 
 	local online_cpus=()
 	local offline_cpus=()
```

PR #42, opened 2018-12-15
Add support for older bash versions

bvanassche:
This patch series allows to run blktests on systems with an older version of bash and that neither have the realpath nor the findmnt executable.

### DIFF
--- a/check
+++ b/check
@@ -4,6 +4,8 @@
 
 shopt -s extglob
 
+. common/realpath
+
 _warning() {
 	echo "$0: $*" >&2
 }
--- a/check
+++ b/check
@@ -24,7 +24,7 @@ _found_test() {
 		return 1
 	fi
 
-	if [[ -z $DESCRIPTION ]]; then
+	if [ -z "${DESCRIPTION:-}" ]; then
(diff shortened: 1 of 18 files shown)


osandov:
What system is this? Bash 4, coreutils with realpath, and util-linux with findmnt all date back to 2010. My usual gauge for what version of utilities to support is Debian's oldstable version, and even that antique has bash 4.3, which was released in 2014. I don't think it's worth the trouble to support anything that outdated.

bvanassche:
The system I'm porting blktests to is the production environment of my employer. That's an environment that really matters to me. The changes I'm proposing make the blktests shell scripts conform better to the shell syntax as defined by the POSIX standard. See also http://pubs.opengroup.org/onlinepubs/9699919799/ and http://pubs.opengroup.org/onlinepubs/9699919799/.

osandov:
Conforming to POSIX shell syntax is a non-goal. Bash has lots of features that make shell scripting less of a headache. If your employer insists on using an environment from last decade, I'd recommend building bash/coreutils/util-linux locally to run these tests.

PR #44, opened 2019-01-15
Multiple srp and nvmeof-mp patches

bvanassche:
This patch series includes several bug fixes for the srp and nvmeof-mp tests and two new srp tests.


tests/srp/014
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0+
+# Copyright (c) 2016-2018 Western Digital Corporation or its affiliates
+
+. tests/srp/rc
+
+RESET_TYPE=$(case "$(basename "$TEST_NAME")" in
```
osandov: This is clever, but can you please just combine these into one test that runs both modes?

bvanassche: I see two possible approaches for combining the two tests into one:
* Issue sg_reset -d and sg_reset -b alternatingly.
* Add a loop such that first sg_reset -d is issued and next sg_reset -b.
I'm not enthusiast about either approach because it will make it harder to figure out whether sg_reset -d or -b made the test fail. Can you explain why you want me to combine the two tests into one?

bvanassche: This comment has been addressed. Please have another look.

bvanassche: Ping?

PR #51, opened 2019-05-01
blktests: common/rc, skip missing modules if they're built into the kernel

jkkm:
Ted complained about needing to have a bunch of things built as
modules instead of built in at LSFMM... fix this for him.

Since we still look at module_param separately, we can skip tests that
can't be unloaded though.

Signed-off-by: Kyle McMartin <jkkm@fb.com>

### DIFF
--- a/common/rc
+++ b/common/rc
@@ -33,6 +33,10 @@ _have_modules() {
 	local module
 
 	for module in "$@"; do
+		# check if a module is already loaded or built in.
+		if test -d "/sys/firmware/$module"; then
+			continue
+		fi
 		if ! modprobe -n -q "$module"; then
 			missing+=("$module")
 		fi


osandov:
AFAICT, modprobe already handles this:

```
$ zgrep CONFIG_EXT4 /proc/config.gz
CONFIG_EXT4_FS=y
CONFIG_EXT4_USE_FOR_EXT2=y
CONFIG_EXT4_FS_POSIX_ACL=y
CONFIG_EXT4_FS_SECURITY=y
# CONFIG_EXT4_DEBUG is not set
$ modprobe -n ext4; echo $?
0
```

So I don't know what exactly was going on with Ted's case.

jkkm:
Hrm, good point, I wonder if he's either using ancient module-init-tools, klibc modprobe, or something else weird is going on.

PR #53, opened 2019-10-15
linux/kvm.h appears to be unnecessary (and is x86 specific)

jrrk:
Thanks for making this code available. I tried to compile your tools on RISCV and found an unnecessary reference to linux/kvm.h

Your comments would be appreciated.

### DIFF
--- a/src/sg/syzkaller1.c
+++ b/src/sg/syzkaller1.c
@@ -34,7 +34,7 @@
 #include <linux/if_ether.h>
 #include <linux/if_tun.h>
 #include <linux/ip.h>
-#include <linux/kvm.h>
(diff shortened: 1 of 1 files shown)


src/sg/syzkaller1.c
```diff
@@ -34,7 +34,7 @@
 #include <linux/if_ether.h>
 #include <linux/if_tun.h>
 #include <linux/ip.h>
-#include <linux/kvm.h>
+//#include <linux/kvm.h>
```
osandov: Yup, this is unnecessary. It's probably syzkaller-generated boilerplate. Could you please update your commit to remove the line entirely instead of commenting it out?

PR #56, opened 2019-10-29
Add --config argument for custom config filenames

andrealmeid:
Instead of just using the default `config` file, one may also find useful to specify which configuration file would like to use without editing the `config` file, like this:

```shell
$ ./check --config=tests_nvme
...
$ ./check -c tests_scsi
```
This pull request solves this. This change means to be optional, in the sense that the default behavior should not be modified and current setups will not be affect by this. To check if this is true, I have done the following test:

- Print the value of variables `$DEVICE_ONLY, $QUICK_RUN, $TIMEOUT, $RUN_ZONED_TESTS, $OUTPUT, $EXCLUDE`
- Run with the following setups:
    - with a `config` file in the dir
    - without a `config` file in the dir
    - configuring using command line arguments

With both original code and with my changes, I validated that the values remained the same. Then, I used the argument `--config=test_config` to check that the values of variables are indeed changing.

This patchset add this feature, update the docs and fix a minor issue with a command line argument. Also, I have changed `# shellcheck disable=SC1091` to `# shellcheck source=/dev/null`, since it seems the proper way to disable this check according to shellcheck documentation[1].

Thanks,
    André

[1] https://github.com/koalaman/shellcheck/wiki/SC1090#exceptions

### DIFF
--- a/Documentation/running-tests.md
+++ b/Documentation/running-tests.md
@@ -21,7 +21,9 @@ will run all tests in the `loop` group and the `block/002` test.
 
 Test configuration goes in the `config` file at the top-level directory of the
 blktests repository. Test configuration options can also be set as environment
-variables instead of in the `config` file.
+variables instead of in the `config` file. It is also possible to use a
+different file for configuration, using `--config=<file_name>`. Note that, for
+multiple uses of this option, all files will be loaded.
 
 ### Test Devices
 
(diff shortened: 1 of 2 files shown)


andrealmeid:
Closing, since it was merged at the mail list: https://patchwork.kernel.org/cover/11220153/

PR #59, opened 2020-04-07
nvme/033: add test case for nvme update hardware queue count

dublio:
Modify nvme module parameter write_queues to change hardware
queue count, then reset nvme controller to reinitialize nvme
with different queue count.

Attention, this test case may trigger a kernel panic.

Signed-off-by: Weiping Zhang <zhangweiping@didiglobal.com>

### DIFF
--- a/tests/nvme/033
+++ b/tests/nvme/033
@@ -1,0 +1,87 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2020 Weiping Zhang <zwp10758@gmail.com>
+#
+# Test nvme update hardware queue count larger than system cpu count
+#
+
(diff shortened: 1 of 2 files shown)


osandov:
Sorry for the delay. Is there a corresponding fix in the kernel for this?

kawasaki:
Today, I ran the proposed test case with kernel v5.18-rc6 and QEMU NVME device and observed test case pass. I'm not sure how to trigger the failure, and not sure if the test case deserves merge to master branch...

kawasaki:
@igaw , thanks for the note. I close this case.
@dublio, it was a bit sorrow that your work did not get response in timely manner. I will try to respond earlier to future pull requests.

PR #64, opened 2020-06-09
Fix ./check: line 275: LAST_TEST_RUN["$key"]: unbound variable

zhijianli88:
for the srp first run where /path/to/blktests/results is empty, it will
throw errors like:

$ ./check srp/001 srp/002 srp/003 srp/004 srp/005 srp/006 srp/007 srp/008 srp/009 srp/010 srp/011 srp/012 srp/013 srp/015
srp/001 (Create and remove LUNs)----------------------------
srp/001 (Create and remove LUNs)                             [failed]

==> /tmp/stderr <==
./check: line 275: LAST_TEST_RUN["$key"]: unbound variable

==> /tmp/stdout <==
    --- tests/srp/001.out>------2020-06-09 13:44:19.000000000 +0800
    +++ /lkp/benchmarks/blktests/results/nodev/srp/001.out.bad>-2020-06-09 16:18:22.594012394 +0800
    @@ -1,3 +1,4 @@
     Configured SRP target driver
    +tests/srp/rc: line 105: use_blk_mq: Permission denied
     count_luns(): 3 <> 3
     Passed

Signed-off-by: Li Zhijian <zhijianx.li@intel.com>

### DIFF
--- a/common/multipath-over-rdma
+++ b/common/multipath-over-rdma
@@ -660,8 +660,6 @@ teardown() {
 setup_test() {
 	local i m modules
 
-	set -u
-
 	if ! shutdown_client; then
 		echo "failed to shutdown client"
 		return 1


bvanassche:
Is this perhaps an alternative for [[PATCH blktests] Restore support for running tests without prior test results](https://lore.kernel.org/linux-block/20200520165241.24798-1-bvanassche@acm.org/)? If so, I like the approach of the that patch better.

zhijianli88:
> Is this perhaps an alternative for [[PATCH blktests] Restore support for running tests without prior test results](https://lore.kernel.org/linux-block/20200520165241.24798-1-bvanassche@acm.org/)? If so, I like the approach of the that patch better.

it doesn't work, since "set -u" means throw when it meets unset variable
```shell
              -u      Treat  unset  variables  and parameters other than the special parameters "@" and "*" as an error when performing parameter expansion.  If expansion is attempted on an unset variable or parameter, the shell prints an error message, and, if not interac‐
                      tive, exits with a non-zero status.
```
beside line 245, others reference this array as well
```
275         for key in "${!TEST_RUN[@]}"; do                                        
276                 if [[ $key =~ ^description|date|status|reason|exit_status$ ]]; then
277                         continue                                                
278                 fi                                                              
279                 # [[ -v array[key] ]] was added in Bash 4.3. Do it this ugly    
280                 # way to support older versions.                                
281                 if [[ -n ${LAST_TEST_RUN["$key"]} || ${LAST_TEST_RUN["$key"]-unset} != unset ]]; then
282                         continue                                                
283                 fi                                                              
284                 value="${TEST_RUN["$key"]}"                                     
285                 printf '    %s\t\t...\t%s\n' "$key" "$value"                    
286         done   
```

osandov:
Thanks, I prefer this fix for now. blktests wasn't written with `set -u` in mind, so setting that for one test group that not everyone runs is going to cause issues.

PR #66, opened 2020-06-16
tests/srp/rc: Do not pass an empty string to dd

bvanassche:
Instead of passing an empty string as argument to dd, do not pass any
argument when not using direct I/O.

Fixes: 577caa7d2b4a ("Fix unquoted integer shellcheck errors")

### DIFF
--- a/tests/srp/rc
+++ b/tests/srp/rc
@@ -516,12 +516,12 @@ start_lio_srpt() {
 	i=0
 	for r in "${vdev_path[@]}"; do
 		if [ -b "$(readlink -f "$r")" ]; then
-			oflag=oflag=direct
+			oflag=(oflag=direct)
 		else
-			oflag=
+			oflag=()
 		fi
 		echo -n "Zero-initializing $r ... " >>"$FULL"
-		dd if=/dev/zero of="${r}" bs=1M count=$((ramdisk_size>>20)) "$oflag" >/dev/null 2>&1 || return $?
+		dd if=/dev/zero of="${r}" bs=1M count=$((ramdisk_size>>20)) "${oflag[@]}" >/dev/null 2>&1 || return $?
 		echo "done" >>"$FULL"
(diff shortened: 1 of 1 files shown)


osandov:
My bad, thanks, Bart.

kawasaki:
@osandov @bvanassche  FYI, after this commit, make check reports a warning in my environment. I think we need to quote "oflag=direct".

> shellcheck -x -e SC2119 -f gcc check new common/* \                                                                    │                                                                                                                 │
>         tests/*/rc tests/*/[0-9]*[0-9]                                                                                 │                                                                                                                 │
> tests/srp/rc:519:11: warning: The = here is literal. To assign by index, use ( [index]=value ) with no spaces. To keep │                                                                                                                 │
> as literal, quote it. [SC2191]                                                                                         │                                                                                                                 │
> make: *** [Makefile:21: check] Error 1

bvanassche:
Is something like the patch below perhaps what you had in mind? Do you want to submit this as a patch or do you perhaps expect me to do that?
~~~
diff --git a/tests/srp/rc b/tests/srp/rc
index 317ef03a3f4f..72a3bca3b44d 100755
--- a/tests/srp/rc
+++ b/tests/srp/rc
@@ -517,7 +517,7 @@ start_lio_srpt() {
 	i=0
 	for r in "${vdev_path[@]}"; do
 		if [ -b "$(readlink -f "$r")" ]; then
-			oflag=(oflag=direct)
+			oflag=("oflag=direct")
 		else
 			oflag=()
 		fi
~~~

kawasaki:
@bvanassche Yes, that change avoids the warning. May I ask you to submit the patch?

bvanassche:
Please take a look at https://github.com/osandov/blktests/pull/72.

kawasaki:
@bvanassche Thank you!

PR #69, opened 2020-06-30
tests: mark tests with CAN_BE_ZONED=1

birkelund:
Zoned devices should have no issues running block/{008,019} and
nvme/032, so mark the tests with CAN_BE_ZONED=1.

Signed-off-by: Klaus Jensen <k.jensen@samsung.com>


osandov:
Thanks! We might also want to rename `_run_fio_rand_io` to `_run_fio_randread` to make it clearer.

PR #81, opened 2021-08-24
Create test name from most recently used test number

jderrick:
The 'new' script can inadvertently use a test name that was removed, ex
nvme/001, which may create confusion if identically-named tests exist
among different versions. Instead, generate a test name at the numerical
tail end of the test group.

Signed-off-by: Jon Derrick <jonathan.derrick@linux.dev>

### DIFF
--- a/new
+++ b/new
@@ -96,13 +96,11 @@ EOF
 	echo "Created tests/${group}/rc"
 fi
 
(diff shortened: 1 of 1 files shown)


osandov:
This seems reasonable. Please fix the missing quoting error from shellcheck and I can pull it in.

PR #84, opened 2021-10-04
Three blktests patches

bvanassche:


tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+config_hz() {
+	if [ -e /proc/config.gz ]; then
+		zcat /proc/config.gz
+	else
+		cat "/boot/config-$(uname -r)"
+	fi | sed -n 's/^CONFIG_HZ=//p'
+}
+
+trigger_scsi_eh() {
+	if ! _init_scsi_debug ""; then
```
osandov, line 31: Why do you need the `""`? Can't this just be `if ! _init_scsi_debug; ...`?

tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+	# Enable SCSI error handler logging
+	echo 63 > /sys/module/scsi_mod/parameters/scsi_logging_level
+
+	local dev freq jdelay target
+	for target in "${SCSI_DEBUG_TARGETS[@]}"; do
+		if ! dev=$(cd "/sys/class/scsi_device/${target}/device/block" &&
+			echo *); then
+			echo "Error: scsi_debug instance not found"
+			continue
+		fi
```
osandov, line 44: You can replace all of this with `for dev in "${SCSI_DEBUG_DEVICES[@]}"`.

tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+		# Change the block layer timeout to max(1 / CONFIG_HZ, 0.001)
+		# seconds.
+		echo 1 > "/sys/class/block/$dev/queue/io_timeout"
+		echo "I/O timeout = $(<"/sys/class/block/$dev/queue/io_timeout")" >>"$FULL"
+		# Change the scsi_debug delay to 5 seconds.
+		freq=$(config_hz)
+		jdelay=$((5 * "${freq}"))
+		echo "CONFIG_HZ=${freq} jdelay=${jdelay}" >>"$FULL"
+		echo "$jdelay" > /sys/module/scsi_debug/parameters/delay
```
osandov, line 53: Can't you use the `ndelay` parameter (in nanoseconds) instead of dealing with jiffies?

tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+
+	echo "Test complete"
+
+	_exit_scsi_debug
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+
+	trigger_scsi_eh
```
osandov, line 80: Just move the contents of `trigger_scsi_eh` into here and get rid of that function.

tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+				echo "Reading from scsi_debug succeeded"
+		  	else
+				echo "Reading from scsi_debug failed"
+			fi
+		} &
+		sleep 5
+		echo "Restoring I/O timeout"
+		# Change the block layer timeout to 1 second.
+		echo 1000 > "/sys/class/block/$dev/queue/io_timeout"
+		wait
```
osandov, line 66: Do you need to do this? We tear down the device soon anyways.

tests/scsi/007
```diff
@@ -0,0 +1,81 @@
...
+
+QUICK=1
+
+requires() {
+	_have_scsi_debug
+}
+
+group_device_requires() {
+	true
+}
```
osandov, line 20: Why is this necessary? This is not a device test.

tests/scsi/007
bvanassche: I will remove the code that changes the timeout to one second.

tests/scsi/007
bvanassche: Done.

tests/scsi/007
bvanassche: Done.

tests/scsi/007
osandov: 2 seconds in nanoseconds does, though. Is that a large enough delay to reliably trigger this?

tests/scsi/007
```diff
@@ -0,0 +1,72 @@
...
+			if dd if="/dev/$dev" of=/dev/null bs=512 count=1 \
+				iflag=direct >&/dev/null; then
+				echo "Reading from scsi_debug succeeded"
+		  	else
+				echo "Reading from scsi_debug failed"
+			fi
+		} &
+		# Give the error handler a chance to run.
+		sleep 5
+		wait
```
osandov, line 63: Now that you're just sleeping and waiting, you might as well run the `dd` in the foreground and get rid of this.

tests/scsi/007
```diff
@@ -0,0 +1,72 @@
...
+
+requires() {
+	_have_scsi_debug
+}
+
+# Override the function with the same name defined in tests/scsi/rc such that
+# this test is run whether or not the test device is a SCSI device.
+group_device_requires() {
+	true
+}
```
osandov, line 22: This test defines the `test()` function, not the `test_device()` function. That means that it is independent of any test devices, so this is unnecessary (see `tests/scsi/004`, which is similar).

osandov (changes requested):

bvanassche:
All review comments should have been addressed. Please let me know if any additional changes need to be made.

osandov:
I'm still not a fan of depending on `CONFIG_HZ`, but that's `scsi_debug`'s fault.

PR #87, opened 2022-03-29
Add I/O scheduler performance tests for queue depth 1

bvanassche:


kawasaki:
I reviewed the three commits and left comments. As I noted in the linu-block list, I think the 2nd patch for QD=1 I/O scheduler test needs to wait for fix patch in the block layer before it merged to blktests master branch not to increase number of failing test cases.

bvanassche:
@kawasaki I reduced this pull request from a patch series into a single patch. Please re-review.

kawasaki:
Hi @bvanassche , thank you for simplifying this PR.
The change looks good. Still I think the new script file for a single function would be too much, but I'm ok with it.

Last question I have is license. The common/iosched has license "GPL-2.0". On the other hand, the default license of blktests is "GPL-3.0+". Do you have specific reason to choose "GPL-2.0" and do not add "+"?  If not, may I ask to change it to "GPL-3.0+"? I would like to keep the blktests license consistent and simple. (I overlooked this point in the past reviews. Sorry about it.)

bvanassche:
The io_schedulers() function has been renamed into _io_schedulers() and has been moved into the file common/rc. Please take another look.

kawasaki:
@bvanassche Thanks for the updates. Applied.

PR #90, opened 2022-05-13
Introduce _dmesg_since_test_start()

zhijianli88:
Prevent previous dmesg confusion

### DIFF
--- a/common/rc
+++ b/common/rc
@@ -333,3 +333,11 @@ unload_module() {
 	done
 	return 1
 }
+
+_have_writeable_kmsg() {
+	if [[ ! -w /dev/kmsg ]]; then
+		SKIP_REASON="cannot write to /dev/kmsg"
+		return 1
+	fi
+	return 0
+}
--- a/check
+++ b/check
@@ -129,6 +129,12 @@ _find_tests() {
 	fi
 }
 
+# Return the dmesg log since the start of this test.
+_dmesg_since_test_start()
+{
+	dmesg | grep -A 9999 "$dmesg_marker"
+}
+
 _check_dmesg() {
 	local dmesg_marker="$1"
 	local seqres="${RESULTS_DIR}/${TEST_NAME}"
@@ -137,7 +143,7 @@ _check_dmesg() {
 		return 0
 	fi
 
-	dmesg | bash -c "$DMESG_FILTER" | grep -A 9999 "$dmesg_marker" >"${seqres}.dmesg"
+	_dmesg_since_test_start | bash -c "$DMESG_FILTER" >"${seqres}.dmesg"
 	grep -q -e "kernel BUG at" \
 	     -e "WARNING:" \
 	     -e "BUG:" \
--- a/tests/meta/rc
+++ b/tests/meta/rc
@@ -18,14 +18,6 @@ group_device_requires() {
 	fi
 }
 
-_have_writeable_kmsg() {
-	if [[ ! -w /dev/kmsg ]]; then
-		SKIP_REASON="cannot write to /dev/kmsg"
-		return 1
-	fi
-	return 0
-}
-
 fake_bug_on() {
 	cat << "EOF" > "$TMPDIR/dmesg"
 ------------[ cut here ]------------
--- a/tests/nvme/003
+++ b/tests/nvme/003
@@ -14,6 +14,7 @@ requires() {
 	_nvme_requires
 	_have_modules loop
 	_require_nvme_trtype_is_fabrics
+	_have_writeable_kmsg
 }
 
 test() {
--- a/tests/nvme/003
+++ b/tests/nvme/003
@@ -35,11 +35,11 @@ test() {
 	# This is ugly but checking for the absence of error messages is ...
 	sleep 10
 
-	if dmesg | grep -q "failed nvme_keep_alive_end_io error="; then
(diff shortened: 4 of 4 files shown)


zhijianli88:
ping

kawasaki:
Hi @zhijianli88 , I think this change has good motivation. I assume you observed false-positive failure of the test case nvme/003, didn't you? Also, please refer the couple of comments I made to the commits.

common/rc
```diff
@@ -333,3 +333,22 @@ unload_module() {
 	done
 	return 1
 }
+
+# Copied from xfstests
```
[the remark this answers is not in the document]
kawasaki: In the "check" script, "grep -A 9999 $dmesg_marker" is used to cut out the dmesg after the marker. I suggest to use same code. This will avoid coping code from xfstests. Could you update the commit to use the grep -A command? (I think it is ok to copy a part of code from xfstests to blktests, but would like to keep such a copy minimum).

kawasaki (changes requested):
Thanks for this pull request. As I commented, I suggest some improvements.

zhijianli88:
> Hi @zhijianli88 , I think this change has good motivation. I assume you observed false-positive failure of the test case nvme/003, didn't you?

Yes, that's true.

common/rc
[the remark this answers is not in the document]
zhijianli88: "grep -A 9999 $dmesg_marker"  sounds good to me.

common/rc
```diff
@@ -333,3 +333,22 @@ unload_module() {
...
+# Return the dmesg log since the start of this test.  Caller must ensure that
+# /dev/kmsg was writable when the test was started so that we can find the
+# beginning of this test's log messages.
+_dmesg_since_test_start()
+{
+	# search the dmesg log of last run of $dmesg_marker
+	dmesg | tac | sed -ne "0,\#$dmesg_marker#p" | tac
+}
+
+my_dmesg()
```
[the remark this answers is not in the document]
zhijianli88: make senses

common/rc
```diff
@@ -333,3 +333,11 @@ unload_module() {
 	done
 	return 1
 }
+
+# Return the dmesg log since the start of this test.
+_dmesg_since_test_start()
+{
+	# make shellcheck happy
+	# shellcheck disable=SC2154
+	dmesg | grep -A 9999 "$dmesg_marker"
```
bvanassche, line 342: This won't work if the dmesg buffer overflowed and hence the `$dmesg_marker` disappeared from the dmesg buffer.

bvanassche (changes requested):

common/rc
zhijianli88: You are right, there is nothing/easy approach we can do in such case, right.

bvanassche: How about changing `dmesg | grep -A 9999 "$dmesg_marker"` into something like the following?
```
dmesg > tmpfile
if grep -q "$dmesg_marker" tmpfile; then
    sed -n "/$dmesg_marker/,\$p" tmpfile
else
    cat tmpfile
fi
```

zhijianli88: IMO, this cannot resolve `dmesg buffer overflowed`.  To avoid the overflow issue, we have to read dmesg through a realtime buffered log service, such as journalctl or /var/log/message or dmesg --follow. To be honest,  currently I don't think we should consider the overflow problem which is a very very low possibility during one test running.

kawasaki: I can agree the concern that @bvanassche, but as @zhijianli88 noted, I think we should not consider these issues within this pull request. I prefer to see other patches or pull requests to cover the issues.

To be more precise, I think two issues are left: 1) dmesg buffer overflow 2) grep -A 9999 commands' 9999 lines limitation. journalctl can be a solution for the issue 1), and the code suggested by @bvanassche will address the issue 2). Anyway, let's not discuss them here :)

common/rc
```diff
@@ -333,3 +333,19 @@ unload_module() {
 	done
 	return 1
 }
+
+# Return the dmesg log since the start of this test.
+_dmesg_since_test_start()
+{
+	# make shellcheck happy
```
kawasaki, line 340: This shellcheck exception is a bit ugly. So I suggest to move this _dmesg_since_test_start function from 'common/rc' to 'check. Probably, just before _check_dmesg() function. On top of that, could you replace the 'dmesg' and 'grep -A 9999' call in _check_dmesg with _dmesg_since_test_start? It will avoid the duplication and future improvement easier.

```

@@ -137,7 +143,7 @@ _check_dmesg() {
                return 0
        fi
 
-       dmesg | bash -c "$DMESG_FILTER" | grep -A 9999 "$dmesg_marker" >"${seqres}.dmesg"
+       _dmesg_since_test_start | bash -c "$DMESG_FILTER" >"${seqres}.dmesg"
        grep -q -e "kernel BUG at" \
             -e "WARNING:" \
             -e "BUG:" \


```

kawasaki (changes requested):
Hi @zhijianli88 
I made one more comment on _dmesg_since_test_start(). Could you rework on it?
Other than that, all changes looks ok for me. Also, I ran blktests in my test environment with your patches and observed no regression. Looks good so far.

common/rc
zhijianli88: sure. i will update it soon

kawasaki:
@zhijianli88 Thanks for fixing up. The commits got merged.

PR #93, opened 2022-06-09
loop: Add test to setup GPT and ESP on raw disk image

michael-schaller:
Add test for setting up a raw disk image with a GPT and ESP.
This is a regression test for commit b9684a71fca7
("block, loop: support partitions without scanning").

Signed-off-by: Michael Schaller <misch@google.com>

### DIFF
--- a/tests/loop/008
+++ b/tests/loop/008
@@ -1,0 +1,46 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2022 Google LLC.
+# Author: misch@google.com (Michael Schaller)
+#
+# Test if partitioning and file system creation on a raw disk image works.
+#
+# The raw disk image is associated to a loop device. The disk image only
+# contains a GPT (GUID Partition Table) and ESP (EFI System Partition).
+#
+# Regression test for commit b9684a71fca7 ("block, loop: support partitions
+# without scanning").
+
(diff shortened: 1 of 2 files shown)


kawasaki (changes requested):
Hi @michael-schaller 
Thank you for this PR. The code logic looks good. Let me make a few comments on other things.

-  Apache license is noted in the file header comment. However, GPL3+ is the license for blktests (some files have GPL2+). To keep license of blktests simple, I wish to see GPL3+ for this new test case also.
- The commit title is rather lengthy. The additional note in parenthesis (regression ...) can be removed, I think.
- The line of commit message is too long. I wish to see 72 characters each in the commit message lines.

Hope these requests are reasonable for you.

michael-schaller:
Done. Please take another look.

kawasaki:
@michael-schaller Thanks for the license change! However, the 2nd commit for license change does not have your Signed-off-by tag, which all blktests commits have. Could you add the tag?

Also, single commit is the better record than the two commits. Could you squash the two commits into one and force push to your git branch? Or if it takes too much time for you, please let me know. I can do that squash work.

michael-schaller:
Done. Please take another look.

kawasaki:
@michael-schaller Thanks for the updates. Applied to the master branch.

michael-schaller:
Thanks. :smiley:

PR #94, opened 2022-06-10
Testsuite for nvme in-band authentication

hreinecke:
This is a testsuite for the upcoming NVMe In-Band authentication. Clearly it depends on the code being merged with upstream, but nevertheless I've prepared the pull request now to give people a chance to test the new code.
nvme-cli has already been updated, so you just need to ensure you're running a newer (>= 2.0) version of nvme-cli.


hreinecke:
Superseded by PR #97.

PR #97, opened 2022-06-29
Testsuite for nvme in-band authentication

hreinecke:
This is a testsuite for the upcoming NVMe In-Band authentication. Clearly it depends on the code being merged with upstream, but nevertheless I've prepared the pull request now to give people a chance to test the new code.
nvme-cli has already been updated, so you just need to ensure you're running a newer (>= 2.0) version of nvme-cli.

It supersedes the previous pull request #94.

PR #98, opened 2022-07-20
Testsuite for nvme in-band authentication

hreinecke:
This is a testsuite for the upcoming NVMe In-Band authentication. Clearly it depends on the code being merged with upstream, but nevertheless I've prepared the pull request now to give people a chance to test the new code.
nvme-cli has already been updated, so you just need to ensure you're running a newer (>= 2.0) version of nvme-cli.

It supersedes the previous pull request https://github.com/osandov/blktests/pull/97.

### DIFF
--- a/tests/nvme/041
+++ b/tests/nvme/041
@@ -1,0 +1,82 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2022 Hannes Reinecke, SUSE Labs
+#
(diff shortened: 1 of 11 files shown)


kawasaki:
@hreinecke Thank you updating the series, and fixing up the SKIP_REASON usage. Actually, the array name should have 'S', SKIP_REASONS. I added another commit to fix it. Anyway, now the tests are is in upstream :)

PR #100, opened 2022-08-30
Enable 'fc' as transport for blktests

hreinecke:
Hi all,

this patchset adds support for the 'fc' transport in blktests by configuring the 'fcloop' device. This allows to validate the nvme-fc transport in addition to the existing 'loop', 'tcp', and 'rdma' transports.

### DIFF
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -343,7 +343,7 @@ _nvme_connect_subsys() {
 	if [[ -n "${ctrlkey}" ]]; then
 		ARGS+=(--dhchap-ctrl-secret="${ctrlkey}")
 	fi
-	nvme connect "${ARGS[@]}"
+	nvme connect "${ARGS[@]}" 2> /dev/null
 }
 
 _nvme_discover() {
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -10,6 +10,10 @@
 def_traddr="127.0.0.1"
 def_adrfam="ipv4"
 def_trsvcid="4420"
+def_remote_wwnn="0x10001100aa000001"
+def_remote_wwpn="0x20001100aa000001"
+def_local_wwnn="0x10001100aa000002"
+def_local_wwpn="0x20001100aa000002"
 def_hostnqn="$(cat /etc/nvme/hostnqn 2> /dev/null)"
 def_hostid="$(cat /etc/nvme/hostid 2> /dev/null)"
 nvme_trtype=${nvme_trtype:-"loop"}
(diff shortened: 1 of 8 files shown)


hreinecke:
Note that 'fcloop' currently has some issues, as it'll retry authentication failures, causing repeated connect/disconnect attempts to fail. Patch is pending in the linux-nvme mailing list ('nvme-auth: retry command if DNR bit is not set').

hreinecke:
Patchset submitted to linux-nvme, thread 'nvme: honor DNR status for authentication' to fix the connect behaviour wrt DNR bit handling.

tests/nvme/rc
```diff
@@ -222,7 +325,11 @@ _nvme_connect_subsys() {
 
 	ARGS=(-t "${trtype}" -n "${subsysnqn}")
 	if [[ "${trtype}" != "loop" ]]; then
-		ARGS+=(-a "${traddr}" -s "${trsvcid}")
+		if [[ "${trtype}" == "fc" ]] ; then
```
kawasaki, line 328: A nit: I think indent level of this part can be shallower:

```
if [[ "${trtype}" == "fc" ]] ; then
  ...
elif [[ "${trtype}" != "loop" ]]; then
  ...
fi
```

tests/nvme/rc
```diff
@@ -242,11 +349,16 @@ _nvme_connect_subsys() {
 _nvme_discover() {
 	local trtype="$1"
 	local traddr="${2:-$def_traddr}"
+	local host_traddr="${3:-$def_host_traddr}"
 	local trsvcid="${3:-$def_trsvcid}"
 
 	ARGS=(-t "${trtype}")
 	if [[ "${trtype}" != "loop" ]]; then
-		ARGS+=(-a "${traddr}" -s "${trsvcid}")
+		if [[ "${trtype}" = "fc" ]]; then
```
kawasaki, line 357: Same nit comment as above.

tests/nvme/rc
```diff
@@ -264,11 +376,16 @@ _create_nvmet_port() {
...
 
+	if [[ "${adrfam}" = "fc" ]]; then
+		trsvcid=
+	fi
 	mkdir "${NVMET_CFS}/ports/${port}"
 	echo "${trtype}" > "${NVMET_CFS}/ports/${port}/addr_trtype"
 	echo "${traddr}" > "${NVMET_CFS}/ports/${port}/addr_traddr"
 	echo "${adrfam}" > "${NVMET_CFS}/ports/${port}/addr_adrfam"
-	echo "${trsvcid}" > "${NVMET_CFS}/ports/${port}/addr_trsvcid"
+	if [[ "${trsvcid}" ]]; then
```
kawasaki, line 386: If we change this condition check as follows, previous hunk is not required, I think:

```
if [[ "${adfram}" != "fc" ]]; then
```

kawasaki (changes requested):
Thanks for the PR. As github CI reported, [shellcheck has some complaints](https://github.com/osandov/blktests/runs/8092181659?check_suite_focus=true). Could you fix them?
Also I made a few minor nit comments. 
I ran NVME test group with nvme_trtype=fc and result looks ok for me.

kawasaki:
@igaw created another PR #108 as v2. I close this PR.

PR #102, opened 2022-09-14
srp/016: Add a test that triggers RDMA hot-unplug

bvanassche:
Kernel commit 8fe4ce5836e9 ("scsi: core: Fix a use-after-free"; v6.0-rc5) fixes a kernel crash triggered by RDMA hot-unplug. This test triggers RDMA hot-unplug and also triggers a kernel crash without the aforementioned kernel commit.

Cc: Li Zhijian <lizhijian@fujitsu.com>

### DIFF
--- a/tests/srp/016
+++ b/tests/srp/016
@@ -1,0 +1,20 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0+
+# Copyright 2022 Google LLC
+
+. tests/srp/rc
+
+DESCRIPTION="RDMA hot-unplug"
+QUICK=1
+
+test_hot_unplug() {
+	use_blk_mq y y || return $?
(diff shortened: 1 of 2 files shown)


kawasaki (changes requested):
Hi @bvanassche , thanks for the PR.
One nit is "DMA" in the commit message. I think you meant "RDMA", didn't you? Please fix it. Or if you want me to fix it at merge, please let me know.

I ran this test case with KASAN enabled kernel v6.0-rc4 on my test machines, but it passed and did not recreate the "Fix use-after-free". I guess the depend on machine configuration or other Kconfig items. If you can think of any condition to recreate the failure, please let me know. Assuming this test case triggers the failure on your environment, I think it's ok to apply this change as it is.

bvanassche:
Thanks for having reported this. This should have been fixed.

PR #107, opened 2023-01-15
Fix multiple shellcheck warnings

bvanassche:
Suppress three shellcheck warnings: two warnings that are not useful and another warning that is reported in contexts where it does not apply.


bvanassche:
@kawasaki , can you please take a look at this pull request?

kawasaki:
The changes look good to me. I've merged this PR. Thanks!

PR #108, opened 2023-01-25
Enable 'fc' as transport for blktests (v2)

igaw:
I've picked up @hreinecke's work started in #100 (I think we can close it). I've address the feedback and fixed also the shellcheck reports (though this was against an older version of this tool)

~~The recently released version of shellcheck is reporting a bunch of new warning (as you are already aware of it: #107). So this PR wont pass the test though yet (waiting for the resolve from #107) Hence I mark this as just as draft.~~

The aim is to get my 'new' test working: https://lore.kernel.org/linux-nvme/20220913065758.134668-1-dwagner@suse.de/

### DIFF
--- a/tests/nvme/041
+++ b/tests/nvme/041
@@ -55,16 +55,14 @@ test() {
 	# Test unauthenticated connection (should fail)
 	echo "Test unauthenticated connection (should fail)"
 	_nvme_connect_subsys "${nvme_trtype}" "${subsys_name}" \
-			     "${def_traddr}" "${def_trsvcid}" \
-			     "${hostnqn}" "${hostid}"
+			     "" "" "${hostnqn}" "${hostid}"
 
 	_nvme_disconnect_subsys "${subsys_name}"
 
 	# Test authenticated connection
 	echo "Test authenticated connection"
 	_nvme_connect_subsys "${nvme_trtype}" "${subsys_name}" \
-			     "${def_traddr}" "${def_trsvcid}" \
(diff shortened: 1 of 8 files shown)


kawasaki:
Hi @igaw , thanks for this v2 PR. Now shellcheck has no complaint. Good. I took a look in the changes, and they look good to me.

I ran nvme test group on kernel v6.2-rc5 with kernel configs NVME_FC=m, NVME_TARGET_FC=m and NVME_TARGET_FCLOOP=m. I saw nvme/003 fails with KASAN nullptr-deref. Do I need to apply some patches to make the nvme test cases pass with fc transport?

igaw:
I did run into the same problems, these tests seem to uncover existing bugs in upstream.

I don't know what the best strategy is from here. Should we try fix all kernel bugs first? Personally I think it's okay that these test fail at this point.  My point is that having these tests in blktests makes it simpler for all developersto run them without the additional step to apply these patches here. Besides, these tests will only run with the additional `nvme_trtype=fc` command line argument. It's not that it will cause trouble (famous last words :))

But it's your call! I am fine either way.

kawasaki:
@igaw  Okay, I agree that it's the better to merge this PR now.  I've merged it. Thank you for taking care of the series :)

igaw:
Thanks! Alright, let's get those kernel bugs addressed as next.

hreinecke:
And you know what, that was precisely the point of this patchset. I _know_ it exposes kernel bugs, but no-one will be looking at it if we don't have an exploiter.

PR #88, opened 2022-04-19
ANA support

hreinecke:
Hi all,

here's a patchset to implement and test ANA support for NVMe.
Nothing special, really, but nice to have as this functionality is use quite heavily in storage arrays and hence should be tested, too.

PR #113, opened 2023-02-27
Add test for ANA support

hreinecke:
This small patchset adds a test for ANA support.

NOTE: there seems to be a bug in the kernel which doesn't align the ANA groups correctly with the paths; the 'out' file displays the output as it _should_ be (and which matches the settings from nvmet). But the output from sysfs seems to have issues matching group ids to paths.
Investigating.


kawasaki:
@hreinecke Thanks for this PR. (And sorry for PR #88. I did not have time to afford for it.)
I will watch out this PR, waiting for the bug fix in the kernel.

PR #115, opened 2023-03-20
Test different queue types for fabrics transports

igaw:
Test if the transport are handling the different queues correctly.

We also issue some I/O to make sure that not just the plain connect
works. For this we have to use a file system which supports direct I/O
and hence we use a device backend.


igaw:
Patches have been merged.

PR #116, opened 2023-03-31
Add missing executable mode

yizhanglinux:
This also remove unused 013.out file


bvanassche:
Why to set +x mode on test scripts? Shouldn't this mode be removed from all test scripts instead?

yizhanglinux:
> 
Not sure. 
Seems it's the default behavior when creating a new blktests test case, and also xfstests

https://github.com/osandov/blktests/blob/676d42c9f932f4b6ca3f25439c2fc09d595ceefd/new#L255
https://github.com/kdave/xfstests/blob/02200cb8853dcd4cb6272126b7f9bc0a83b889c6/new#L186

kawasaki:
@yizhanglinux I've merged the patches. Thanks!

PR #117, opened 2023-04-05
test queue count changes on reconnect

igaw:
The target is allowed to change the number of i/o queues. test if the host is able to reconnect in this scenario.

Note this is based on top of my previous PR #115, which should go in first.


igaw:
Patches have been merged.

PR #118, opened 2023-05-04
src/Makefile: forward LDFLAGS to $(CC)

Ablu:
Something external may set some special LDFLAGS. Currently those would get ignored. Lets honor them and pass them to the compiler.

This simplifies packaging and cross-compilation.


kawasaki:
@Ablu I've merged this PR. Thanks!

PR #119, opened 2023-05-05
test/nvme: add missing _have_fio check for nvme/045

yizhanglinux:


kawasaki:
@yizhanglinux Applied, thanks!

PR #121, opened 2023-06-16
nvme/rc: Avoid triggering host nvme-cli autoconnect

igaw:
When the host has enabled the udev/systemd autoconnect services for the fc transport it interacts with blktests and make tests break.

nvme-cli learned to ignore connects attemps when using the --context command line option paired with a volatile configuration. Thus we can mark all the resources created by blktests and avoid any interaction with the systemd autoconnect scripts.

### DIFF
--- a/tests/nvme/003
+++ b/tests/nvme/003
@@ -29,8 +29,8 @@ test() {
 
 	loop_dev="$(losetup -f)"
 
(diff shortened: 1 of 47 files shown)


igaw:
This series has been directly merged.

PR #124, opened 2023-07-26
common/scsi_debug: Use _unload_module() instead of modprobe -r

bvanassche:


kawasaki:
@bvanassche Thanks. I'll take a look in the patches next week.

kawasaki:
Have applied. Thanks!

PR #127, opened 2023-10-24
Allow for pre-defined UUID and subsystem NQN

hreinecke:
When running against a pre-defined target we might not be able to set the UUID/WWID, and the subsystem NQN will not be the default subsystem NQN. This patchset avoid to register either of those in the 'out' files such that we can attempt to run against external targets.

### DIFF
--- a/tests/nvme/003.out
+++ b/tests/nvme/003.out
@@ -1,3 +1,2 @@
 Running nvme/003
-NQN:nqn.2014-08.org.nvmexpress.discovery disconnected 1 controller(s)
 Test complete
--- a/tests/nvme/004
+++ b/tests/nvme/004
@@ -29,8 +29,7 @@ test() {
 
 	local nvmedev
 	nvmedev=$(_find_nvme_dev "${def_subsysnqn}")
-	cat "/sys/block/${nvmedev}n1/uuid"
-	cat "/sys/block/${nvmedev}n1/wwid"
+	_check_uuid "${nvmedev}"
 
 	_nvme_disconnect_subsys ${def_subsysnqn}
 
--- a/tests/nvme/004.out
+++ b/tests/nvme/004.out
@@ -1,5 +1,3 @@
 Running nvme/004
-91fdba0d-f87b-4c25-b80f-db7be1418b9e
-uuid.91fdba0d-f87b-4c25-b80f-db7be1418b9e
 NQN:blktests-subsystem-1 disconnected 1 controller(s)
 Test complete
--- a/tests/nvme/004.out
+++ b/tests/nvme/004.out
@@ -1,3 +1,2 @@
 Running nvme/004
-NQN:blktests-subsystem-1 disconnected 1 controller(s)
 Test complete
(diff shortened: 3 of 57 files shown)


kawasaki:
@hreinecke Thanks for the patches. I think these are good preparation to introduce nvme test runs with real RDMA hardware.

I have a couple of comments:

1. As for the patch "nvme: Do not print UUID to log files", _check_uuid() just confirms UUID and wwid are consistent. How about to compare UUID against def_subsys_uuid also for more robust check? Here I assume that def_subsys_uuid will hold UUID of the real RDMA setup, but this may not be true.
2. As for the patch "nvme: do not print subsystem NQN to stdout", it removed all checks for outputs of 'nvme disconnect' command. I think this may reduce the ability to catch unexpected kernel behavior. How about just remove the NQN part of 'nvme disconnect' output? Still left part of the output like "disconnected 1 controller(s)" can be used to catch something.

Based on these ideas, I implemented two commits on top of your patches:
https://github.com/kawasaki/blktests/commit/07fc073cd3d8484289939392a01bf59133f21998
https://github.com/kawasaki/blktests/commit/d551bcdb9c27ccd3ef656557363e8355fb2f55d9

igaw:
About point 2). I'd propose we check the state of the controller after `nvme connect` and `nvme disconnect` instead parsing the output of `nvme`.

I've done this for the new tests `nvme/050`:

https://lore.kernel.org/linux-nvme/20230621155825.20146-2-dwagner@suse.de/

kawasaki:
> About point 2). I'd propose we check the state of the controller after `nvme connect` and `nvme disconnect` instead parsing the output of `nvme`.
> 
> I've done this for the new tests `nvme/050`:
> 
> https://lore.kernel.org/linux-nvme/20230621155825.20146-2-dwagner@suse.de/

@igaw Thanks for the comment. I found that nvme/048 does similar thing as the patch for nvme/050. This state check approach should work for `nvme connect`, but I'm not sure if it works for `nvme disconnect`. After disconnect success, the nvme device and the sysfs state file are gone. What we can do is just checking if the those files are left or not.

I do not stick to the point 2). If there is no push back on linux-nvme, my comment should not block this work.

igaw:
Okay, I don't really mind. Anyway, these patches look good to me.

kawasaki:
I will try to post the patches from @hreinecke to linux-nvme list next week. I will squash my patches, assuming Hannes is ok with them.
(I will take vacation from tomorrow, so my response will be slow.)

hreinecke:
Tend to agree with @igaw . There is value in figuring out whether 'nvme connect' or 'nvme disconnect' succeeded, and we should register that.
Ideally the script should bail out here, ie 'nvme connect' / 'nvme disconnect' should return an error code, and the script should terminate if an error is registered.
But that would require proper cleanup functions via 'trap', and would require a larger rework.
Cool project, mind. But for another day.

kawasaki:
> Tend to agree with @igaw . There is value in figuring out whether 'nvme connect' or 'nvme disconnect' succeeded, and we should register that.
> Ideally the script should bail out here, ie 'nvme connect' / 'nvme disconnect' should return an error code, and the script should terminate if an error is registered.
> But that would require proper cleanup functions via 'trap', and would require a larger rework.
> Cool project, mind. But for another day.

@hreinecke Thanks for the comments. I agree to separate the work for the error handling improvement.

I have posted your patches to linux-nvme and linux-block in case anyone has comments. As for my suggestions, I reflected only my suggestion for the UUID patch. Let's wait and see comments on the list for a week or so.

PR #131, opened 2024-01-10
nvme/rc: don't print the nvme connect msg

yizhanglinux:
With commit [1], the nvme connect command will show the connect msg which breaks blktests nvme/ related cases[2], fix it from blktests side.

[1]
https://github.com/linux-nvme/nvme-cli/commit/0b8d1e03049c5092d705bcd3ce369f02a9472f95 
[2]
$ ./check nvme/003
nvme/003 (test if we're sending keep-alives to a discovery controller) [failed]
    runtime  11.344s  ...  11.346s
    --- tests/nvme/003.out	2024-01-10 04:03:21.975035862 +0100
    +++ /root/blktests/results/nodev/nvme/003.out.bad	2024-01-10 07:19:46.978193215 +0100
    @@ -1,3 +1,4 @@
     Running nvme/003
    +connecting to device: nvme0
     disconnected 1 controller(s)
     Test complete

### DIFF
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -576,7 +576,7 @@ _nvme_connect_subsys() {
 		ARGS+=(--ctrl-loss-tmo="${ctrl_loss_tmo}")
 	fi
 
-	nvme connect "${ARGS[@]}" 2> /dev/null
+	nvme connect "${ARGS[@]}" 2> /dev/null | grep -v "connecting to device:"
 
 	# Wait until device file and uuid/wwid sysfs attributes get ready for
 	# all namespaces.


kawasaki:
@yizhanglinux

Thanks for catching this. Your change will avoid the issue, but it will throw away all of the nvme connect command outputs. So following line would be the better, IMO.
```
nvme connect "${ARGS[@]}" 2> /dev/null | grep -v "connecting to device:"
```

@igaw 

Just in case you want to, feel free to chime in.

yizhanglinux:
> @yizhanglinux
> 
> Thanks for catching this. Your change will avoid the issue, but it will throw away all of the nvme connect command outputs. So following line would be the better, IMO.

Thanks for the review, just updated it as you suggested.

> 
> ```
> nvme connect "${ARGS[@]}" 2> /dev/null | grep -v "connecting to device:"
> ```
> 
> @igaw
> 
> Just in case you want to, feel free to chime in.

igaw:
I think it's better to check for a specific condition (or string in this case) to present or absent. This means in this case we would expect to see the string `connected to device` and fail it is not there. 

But then this would make blktests depended on a specific version of nvme-cli. On the other hand if we use `nvme connect -o normal` it should also trigger the same problem with all nvme-cli 2.x versions, thus we could make the check to verify the `connected to device` presence. Just as idea.

Either way I am fine with the suggested fix. We can revisit this if we want to extend the checks, e.g. are we really connected connected by looking directly on the sysfs states instead relying on nvme-cli.

kawasaki:
@igaw Thanks for the comments and ideas. I tried your idea and created a patch as follows. It looks working, but at the same time, it looks too much for me. I noticed that nvme-cli v2.7 changed the string from "device :" to "connecting to device :", so it makes this patch a bit more confusing. So, I will take @yizhanglinux 's latest commit. I agree that we can revisit the nvme commands result check improvement later.

```
diff --git a/tests/nvme/rc b/tests/nvme/rc
index b0ef1ee..56730bc 100644
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -468,6 +468,7 @@ _nvme_connect_subsys() {
        local reconnect_delay=""
        local ctrl_loss_tmo=""
        local no_wait=false
+       local connect_out
        local i
 
        while [[ $# -gt 0 ]]; do
@@ -576,7 +577,13 @@ _nvme_connect_subsys() {
                ARGS+=(--ctrl-loss-tmo="${ctrl_loss_tmo}")
        fi
 
-       nvme connect "${ARGS[@]}" 2> /dev/null
+       connect_out=$(nvme connect -o normal "${ARGS[@]}")
+       # Confirm connection success by nvme command output
+       if ! grep -q "device:" <<< "${connect_out}"; then
+               echo "nvme connect did not report device connection"
+       fi
+       # Print outputs for comparison except device unique name
+       grep -v "device:" <<< "${connect_out}"
 
        # Wait until device file and uuid/wwid sysfs attributes get ready for
        # all namespaces.
```

PR #138, opened 2024-04-16
block/033 fix

yizhanglinux:


tests/block/033
```diff
@@ -7,7 +7,7 @@
 # can be covered
 
 . tests/block/rc
-. common/ublk
+. tests/ublk/rc
```
kawasaki, line 10: This creates dependency from tests/block to tests/ublk, which is not good. Instead, I suggest to take two actions:
1) move the UBLK_PROG declarations below from tests/ublk/rc to common/ublk
```
if which rublk > /dev/null 2>&1; then
       export UBLK_PROG="src/rublk_wrapper.sh"
else
       export UBLK_PROG="src/miniublk"
fi
```
2) Replace ublk_prog in tests/block/033 with UBLK_PROG.

kawasaki:
Hi @yizhanglinux , thanks for this PR.
As I commented, I suggest some modifications to the first commit. It might be the better to separate it into another PR, and ping @ming1 (Ming Lei) to review.
The other two patches look good to me. The commit for nvme group conflict with my "repeat by conditions" series, but it should be easy to resolve.

tests/block/033
yizhanglinux: Thanks updated as suggested, will open another PR for the misc fix.

yizhanglinux: [@ming1](https://github.com/ming1)  Please help review this change, thanks.

kawasaki:
@yizhanglinux Thanks! This PR has got merged. FYI, I made a follow-up commit for additional fixes.

PR #144, opened 2024-07-16
Patch fixing test case issue which use to failed in old kernel , this…

PraveenPenguin:
… patch skip

test case if test runs in older kenrel less than 6.11

log:

[root@linux blktests]#  ./check loop/010
loop/010 (check stale loop partition)                        [not run]
    Kernel version too old

### DIFF
--- a/tests/loop/010
+++ b/tests/loop/010
@@ -11,6 +11,7 @@ DESCRIPTION="check stale loop partition"
 TIMED=1
 
 requires() {
+        _have_kver 6 11 
 	_have_program parted
 	_have_program mkfs.xfs
 }


disgoel (approved):
Thanks for the fix @PraveenPenguin. This looks good.
```
# ./check loop/010
loop/010 (check stale loop partition)                        [not run]
    runtime  91.128s  ...
    Kernel version too old
```

kawasaki:
I checked the commit, and found that three points need improvements:

1. Spaces were used for indent instead of tabs.
2. There was a trailing space at the modified line end.
3. The commit message had room to improve, such as title format, miss-spelling, or line character count.

I have done manual edits to address them, and applied it manually. Now the commit is in the master branch.
Thank you for the contribution!

PraveenPenguin:
thanks @kawasaki

PR #145, opened 2024-10-01
block/031: remove the fio verify.state file

yizhanglinux:
Remove the fio verify.state file local-block-031-0-verify.state after fio finished

### DIFF
--- a/common/fio
+++ b/common/fio
@@ -198,8 +198,7 @@ _run_fio_rand_io() {
 
 _run_fio_verify_io() {
 	_run_fio --name=verify --rw=randwrite --direct=1 --ioengine=libaio --bs=4k \
-		--iodepth=16 --verify=crc32c "$@"
-	rm -f local*verify*state
+		--iodepth=16 --verify=crc32c --verify_state_save=0 "$@"
 }
 
 _fio_perf_report() {
--- a/common/multipath-over-rdma
+++ b/common/multipath-over-rdma
@@ -564,6 +564,7 @@ run_fio() {
 		grep -q ', io=[0-9].*, run=[0-9]' "$output"
 		rc=$?
 	fi
+	rm -fr ${fio_aux_path}
 	echo "run_fio exit code: $rc" >>"$FULL"
 	return $rc
 }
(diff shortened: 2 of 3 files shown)


kawasaki:
@yizhanglinux Thank you for catching this! The file should not be left.

As for the solution, how about to add "--verify_state_save=0" option to the fio command? This approach does not create the file, then a bit cleaner than "create the file and remove it".

```
diff --git a/tests/block/031 b/tests/block/031
index 5ad6d88..99615ec 100755
--- a/tests/block/031
+++ b/tests/block/031
@@ -45,7 +45,7 @@ test() {
                --ioengine=libaio --runtime="${TIMEOUT}" --thread \
                --name=block-031 --filename=/dev/nullb1 \
                --output="${RESULTS_DIR}/block/fio-output-031.txt" \
-               >>"$FULL"
+               --verify_state_save=0 >>"$FULL"
        fio_status=$?
        rmdir /sys/kernel/config/nullb/nullb1
        _exit_null_blk
```

yizhanglinux:
> @yizhanglinux Thank you for catching this! The file should not be left.
> 
> As for the solution, how about to add "--verify_state_save=0" option to the fio command? This approach does not create the file, then a bit cleaner than "create the file and remove it".
> 
> ```
> diff --git a/tests/block/031 b/tests/block/031
> index 5ad6d88..99615ec 100755
> --- a/tests/block/031
> +++ b/tests/block/031
> @@ -45,7 +45,7 @@ test() {
>                 --ioengine=libaio --runtime="${TIMEOUT}" --thread \
>                 --name=block-031 --filename=/dev/nullb1 \
>                 --output="${RESULTS_DIR}/block/fio-output-031.txt" \
> -               >>"$FULL"
> +               --verify_state_save=0 >>"$FULL"
>         fio_status=$?
>         rmdir /sys/kernel/config/nullb/nullb1
>         _exit_null_blk
> ```

Yeah, that's a better way, just pushed another two commit for the verify.state file cleanup.

yizhanglinux:
++ @bvanassche

kawasaki:
@yizhanglinux thanks, all three changes look good to me. I ran whole blktests, and did not see regression. One small nit change is the commit message of the 3rd commit. To fix it, I applied the three commits manually. Now they are in the master branch. Thank you!

PR #146, opened 2024-10-02
Implement ANA support

hreinecke:
Hi all,
here's a patchset to implement ANA support, together with a basic test for ANA failover while I/O is running.
As usual, comments and reviews are welcome.

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -423,10 +423,11 @@ _nvme_ns_ready() {
 }
 
 _create_nvmet_port() {
-	local trtype="$1"
+	local trtype="${1:-$nvme_trtype}"
 	local traddr="${2:-$def_traddr}"
 	local adrfam="${3:-$def_adrfam}"
 	local trsvcid="${4:-$def_trsvcid}"
+	local portcfs
 
 	local port
 	for ((port = 0; ; port++)); do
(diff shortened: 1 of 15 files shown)


kawasaki:
@hreinecke Thank you for this pull request. I made some comments using this GitHub review system. When I ran nvme test group with the changes, many test cases failed. My comments are trying to avoid those failures. Please check and see if each comment makes sense.

I wondered if the added test case nvme/054 should be skipped if the kernel does not support ANA. However, I noticed that the oldest long term branch 4.19.y supports ANA, so there is no need to check if the kernel supports ANA.

kawasaki:
Today I looked back the series, and noticed that "local -n" was used before this patch series. That's why @hreinecke used "local -n" to simplify the scripts. This means that the existing "local -n"  did not work with bash version 4.2, but nobody complained about it. It might be the time to bump up the bash version requirement for blktests from 4.2 to 4.3.

I'm not yet sure if do that bash requirement version bump. Now I'm trying to remove the "local -n" by myself to see how the script will look like. Tomorrow, I will share my work, hopefully.

kawasaki:
@hreinecke I spent some time to improve your series, and pushed the outcome to [the ana-rework branch in my repo](https://github.com/kawasaki/blktests/commits/ana-rework/). I dropped one of your patches titled "nvme: sanitize _nvme_get_ctrl_list", and added 7 commits to reflect my comments.

Through this work, I found that "local -n" nameref can be avoided with rather clean way. So now I think we should keep the bash version 4.2 requirement. I also confirmed that the all test cases in nvme test groups pass with tranports loop, rdma and tcp. 

Feel free to squash my 7 commits to your commits, if you think they are reasonable.

hreinecke:
I've now included your changes (especially the 'read -ra' trick, very nice!), and added a few more cleanups which I had pending. Please check.

kawasaki:
Hi @hreinecke , I took a look in the branch. IMO, all changes have good motivations. Still I observe some test run failures and a shellcheck warning, but they do not look a big problems. So I hope these changes settled in the blktests.

On the other hand, I think this PR is too big to apply to the master branch in one shot. It has 26 patches. Also, I think it's the better to get reviews on other Linux nvme experts. For these two reasons, I suggest to separate the 26 patches into 3 or 4 series and post to linux-nvme list one by one for review. This will take some time (I guess one week or two for each series, then in total one month or so), but will help to catch some missing mistakes, and will not surprise blktests users.

If you agree, I would like to drive this multi-series cycle. Based on your branch, I will repeat the steps to, create a series, post to linux-nvme, and reflect review comments and apply them to the blktests master. Or if you want to do it by yourself, it is perfectly okay for me. Please let me know your thoughts.

hreinecke:
It's perfectly fine to split it into a multi-series patchset. Please go ahead.

kawasaki:
I have just applied two other patches by Guixin, which conflict with this PR. Yesterday, I spent sometime to resolve the conflicts, and pushed it to [here](https://github.com/kawasaki/blktests/tree/ana-rework). Will evaluate them and split into a few series to post to linux-block.

kawasaki:
I split the patches into three series. The first series was applied to the blktests master reflecting review comments on the list, from the commit e3fe06e to fe6cd13. Now I have just posted [the second series](https://lore.kernel.org/linux-nvme/20241107111354.80606-1-shinichiro.kawasaki@wdc.com/) for the review.

kawasaki:
I [applied the 3rd series](https://lore.kernel.org/linux-nvme/nmvxcnkntu62jmelkyjhnqs3swhbqf4hcavmdd4nr4ruglrsf2@nk4r2xbrwxba/). The work is done :) Thanks again, @hreinecke

PR #151, opened 2024-11-29
nvme/029: capture all the FULL log from test_user_io

yizhanglinux:


kawasaki:
@yizhanglinux   I merged the commit. Thanks!
@ming1 FYI.

PR #152, opened 2024-12-09
nvme/049: update fio_output msg check for io_uring_cmd support

yizhanglinux:
Update the error info check string:
check: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=io_uring_cmd, iodepth=1 fio-3.38-16-g6f3de
Starting 1 process
fio: pid=9638, err=1/file:engines/io_uring.c:1272, func=io_queue_init, error=Operation not permitted

Run status group 0 (all jobs):

### DIFF
--- a/tests/nvme/049
+++ b/tests/nvme/049
@@ -37,7 +37,8 @@ test_device() {
 	# check security permission
 	if ! fio_output=$(fio --name=check --size=4k --filename="$ngdev" \
 			    --rw=read --ioengine=io_uring_cmd 2>&1) &&
-			grep -qe "Permission denied" <<< "$fio_output"; then
+			grep -q -e "Operation not permitted" \
+				-e "Permission denied" <<< "$fio_output"; then
 		SKIP_REASONS+=("IORING_OP_URING_CMD is not allowed for $ngdev")
 		return
 	fi


kawasaki:
The patch changes to the error string to check from "Permission Denied" (EACCES) to "Operation not permitted" (EPERM). As recorded in [the past discussion](https://lore.kernel.org/linux-nvme/p7rz7dz6t2vdihusunblbw6tm4cqrq4vcfaupttjwqqj7jiwgc@6emzgswzmmeq/), I suggested to add the check for EACCES to confirm that SELinux allows io_uring_cmd. So I think still we need the EACCES check.

Then I suggest to check both EACCES and EPERM as below.

```
diff --git a/tests/nvme/049 b/tests/nvme/049
index 599ab58..f25cba8 100755
--- a/tests/nvme/049
+++ b/tests/nvme/049
@@ -37,7 +37,8 @@ test_device() {
        # check security permission
        if ! fio_output=$(fio --name=check --size=4k --filename="$ngdev" \
                            --rw=read --ioengine=io_uring_cmd 2>&1) &&
-                       grep -qe "Permission denied" <<< "$fio_output"; then
+                       grep -q -e "Operation not permitted" \
+                             -e "Permission denied" <<< "$fio_output"; then
                SKIP_REASONS+=("IORING_OP_URING_CMD is not allowed for $ngdev")
                return
        fi
```

One point I wonder is why the EPERM is observed. Regardless of this question, I think still the check for both EACCES and EPERM will be a good change.

yizhanglinux:
Here is the 6.13 which I download from and reproduce environment, in case you want to have a try.

[1]
https://kojipkgs.fedoraproject.org//packages/kernel/6.13.0/
[2]
```
[root@storageqe-40 blktests]# uname -r
6.13.0-0.rc2.22.eln144.x86_64
[root@storageqe-40 blktests]# fio --version
fio-3.36
[root@storageqe-40 blktests]# cat /proc/sys/kernel/io_uring_disabled
2
[root@storageqe-40 blktests]# fio --name=check --size=4K --filename=/dev/ng0n1 --rw=rw --ioengine=io_uring_cmd
check: (g=0): rw=rw, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=io_uring_cmd, iodepth=1
fio-3.36
Starting 1 process
fio: pid=5601, err=1/file:engines/io_uring.c:1090, func=io_queue_init, error=Operation not permitted


Run status group 0 (all jobs):

```

yizhanglinux:
Will update the patch to inlcude the two strings.

kawasaki:
@yizhanglinux Thanks for the update. I've merged it.

kawasaki:
> [root@storageqe-40 blktests]# cat /proc/sys/kernel/io_uring_disabled
> 2

Now I see when fio reports "Operation not permitted". With the merged fix, the test case should be skipped when io uring is disabled. I guess introducing _io_uring_enable/restore will make the test case run when io uring is disabled. But this is beyond the scope of this PR.

PR #154, opened 2024-12-10
zbd/012: Test requeuing of zoned writes and queue freezing

bvanassche:
Test concurrent requeuing of zoned writes and request queue freezing. While this test passes with kernel 6.9, it triggers a hang with kernels 6.10..6.12. This shows that this hang is a regression introduced by the zone write plugging code.

sysrq: Show Blocked State
task:(udev-worker)   state:D stack:0     pid:75392 tgid:75392 ppid:2178   flags:0x00000006
Call Trace:
 <TASK>
 __schedule+0x3e8/0x1410
 schedule+0x27/0xf0
 blk_mq_freeze_queue_wait+0x6f/0xa0
 queue_attr_store+0x60/0xc0
 kernfs_fop_write_iter+0x13e/0x1f0
 vfs_write+0x25b/0x420
 ksys_write+0x65/0xe0
 do_syscall_64+0x82/0x160
 entry_SYSCALL_64_after_hwframe+0x76/0x7e

### DIFF
--- a/tests/zbd/012
+++ b/tests/zbd/012
@@ -1,0 +1,78 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0+
+# Copyright (C) 2024 Google LLC
+#
+# Test concurrent requeuing of zoned writes and request queue freezing. It
+# triggers a hang in the kernel 6.10..6.12 zone write plugging implementation.
+
+. tests/zbd/rc
+. common/scsi_debug
+
+DESCRIPTION="test requeuing of zoned writes and queue freezing"
+TIMED=1
+
+requires() {
+	_have_fio_zbd_zonemode
+}
+
+toggle_iosched() {
+	local iosched
+
+	while true; do
+		for iosched in none mq-deadline; do
+			echo "${iosched}" > "/sys/class/block/$(basename "$zdev")/queue/scheduler"
+			sleep .1
+		done
+	done
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+
+	# To do: make this test pass for qd > 1.
+	local qd=1
+	echo "$qd"
+	local scsi_debug_params=(
+		delay=0
+		dev_size_mb=1024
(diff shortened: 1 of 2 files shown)


bvanassche:
@kawasaki @damien-lemoal Please take a look.

damien-lemoal:
Bart,

The zone write plugging fixes that I posted and that Jens applied should fix this. Please try again with these patches to check.

bvanassche:
Thanks for having taken a look Damien. I have already verified that your patches fix the reported issue. Since the kernel versions referred to above have been released and cannot be modified anymore, I think the references to kernel versions above are correct.

damien-lemoal:
Yes. The fixes are marked for stable so they will be backported to stable 6.12. 6.10 and 6.11 are already EOL so they will not get the backports.

bvanassche:
With "6.10..6.12" I want to refer to kernel versions 6.10.0, 6.11.0 and 6.12.0. In other words, I want to exclude later versions that include backports. Please let me know if you want me to include this clarification in the patch description.

kawasaki:
@bvanassche Thanks for this PR. The code looks good to me. I will wait for the kernel side fix gets upstreamed (v6.13-rc3 or rc4, hopefully?) then merge this PR.

bvanassche:
I think that Linus just pulled Damien's kernel patches :-) See also https://lore.kernel.org/linux-block/d2acd3cb-188c-4ad5-91db-efbc6e50a1c1@kernel.dk/.

kawasaki:
Yes, the fix is now upstreamed and tagged with v6.13-rc3. I have merged this PR. Thanks!

yizhanglinux:
@kawasaki  @bvanassche   To run this case, seems we also need _io_uring_enable/_io_uring_restore for the fio tests.

bvanassche:
I will switch to libaio since io_uring does not preserve the write submission order.

PR #156, opened 2024-12-17
zbd/012: Test higher queue depths

bvanassche:
Instead of only testing queue depth 1, also test higher queue depths.
Use libaio for higher queue depths since io_uring does not preserve the
write order if qd >= 2 and a request queue is frozen.

### DIFF
--- a/tests/zbd/012
+++ b/tests/zbd/012
@@ -29,46 +29,53 @@ toggle_iosched() {
 test() {
 	echo "Running ${TEST_NAME}"
 
-	# To do: make this test pass for qd > 1.
-	local qd=1
-	echo "$qd"
-	local scsi_debug_params=(
-		delay=0
-		dev_size_mb=1024
-		every_nth=$((2 * qd))
-		max_queue="${qd}"
-		opts=0x8000          # SDEBUG_OPT_HOST_BUSY
-		sector_size=4096
-		zbc=host-managed
-		zone_nr_conv=0
-		zone_size_mb=4
-	)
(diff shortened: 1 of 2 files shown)


bvanassche:
Cc-ing @damien-lemoal in case you would like to take a look.

kawasaki:
Thanks Bart. I found [the relevant discussion in the linux-block list](https://lkml.kernel.org/linux-block/e75812ec-9b91-42d0-9ca5-d4bae031e319@acm.org/). I still not sure if it is agreed that libaio should be used instead of i_uring for QD>2 case. Will wait hoping the dicussion gets concluded.

bvanassche:
From https://lkml.kernel.org/linux-block/20241218065859.GA25215@lst.de/:
```
On Tue, Dec 17, 2024 at 12:59:00PM -0800, Damien Le Moal wrote:
> Thanks for the pointer. Will have a look. It may be as simple as always using
> the io-wq worker for zone writes and have these ordered (__WQ_ORDERED). Maybe.

No.  Don't force ordering on people for absolutely no reason.  Use the
new append writes that return the written offsets we've talked about.
Enforcing ordering on I/O submissions/commands/requests always ends up
badly.
```
Does this mean that io_uring won't support preserving the I/O submission order in the near future? Does that answer mean that there won't be a good alternative to libaio for zoned writes with QD > 1? Is this sufficient to have another look at this pull request?

damien-lemoal:
Yes and yes is my answer to your first 2 questions. So I think that switching the test to using libaio is enough and makes the test reliable and valid.

PR #160, opened 2025-02-19
fix(tests/nvme/056): Update ynl path

nuclearcat:
The ynl Python code has been moved to a dedicated directory in Linux kernel commit ab88c2b3739a. Update the test to reflect this new location.

### DIFF
--- a/tests/nvme/056
+++ b/tests/nvme/056
@@ -38,15 +38,15 @@ requires() {
 
 have_netlink_cli() {
 	local cli
-	cli="${KERNELSRC}/tools/net/ynl/cli.py"
+	cli="${KERNELSRC}/tools/net/ynl/pyynl/cli.py"
 
 	if ! [ -f "$cli" ]; then
-		SKIP_REASONS+=("Kernel sources do not have tools/net/ynl/cli.py")
+		SKIP_REASONS+=("Kernel sources do not have tools/net/ynl/pyynl/cli.py")
 		return 1
 	fi
 
 	if ! "$cli" -h &> /dev/null; then
-		SKIP_REASONS+=("Cannot run the kernel tools/net/ynl/cli.py")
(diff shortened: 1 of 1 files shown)


kawasaki:
@nuclearcat Thank you for catching this!

The kernel commit which moved the python scripts is in v6.14-rc1 kernel. Can we make the test case work with the kernels older than v6.14-rc1 also? This will help to users who want to run the test case with v6.13.X kernels.

nuclearcat:
Thank you for your suggestion about supporting older kernels for this test case.
I'd like to clarify that the feature being tested is still pending inclusion in the kernel as patches in netdev and hasn't been merged yet. Since the functionality doesn't exist in any released kernel version (including v6.13.X or v6.14-rc1), the test will fail regardless of script location or kernel version.
The test case is being prepared in anticipation of the feature being merged soon. Once the patches are accepted, the test will begin working with whichever kernel version includes the merged feature.
Adding compatibility with older kernels wouldn't be helpful in this particular case, as the test is specifically verifying functionality that simply doesn't exist yet in any released kernel.
I can submit a separate PR once the feature is included in the kernel, which will add proper version checking to verify the minimal kernel version required. This would ensure the test gracefully handles execution on kernels that don't support the feature.

kawasaki:
Ah, thanks for the clarification. I understand that there is no need to support the older kernel versions. I'll wait for the key kernel commit will be merged. I'm ok both to reuse this PR, or to create a separate PR.

Let me leave a comment about the kernel version check. When a blktests test case requires a specific kernel feature, it is not ideal to refer to the kernel version to check if the requirement is fulfilled, since such check is fragile. It is the better to check something visible from userland for such check: sysfs attribute file existence/value or module parameter existence check. Some test cases refer to kernel version with _have_kver() function, but it is the last resort.

Thanks!

PR #162, opened 2025-03-05
common/nvme: use correct local wwnn/wwpn for rport removal

igaw:
The local wwnn and wwpn don't need a port offset.

WARNING: Test did not clean up port: 0
common/nvme: line 172: _local_wwnn: command not found
common/nvme: line 172: _local_wwpn: command not found
WARNING: Test did not clean up subsystem: blktests-subsystem-1


kawasaki:
@igaw I've merged this PR. Thanks!

PR #163, opened 2025-03-14
src/Makefile: include CFLAGS in probe macros

stsquad:
When cross-building we need to pass the correct paths via CFLAGS to be able to find extra libraries. We could convert to using pkg-config to resolve library availability but as the rest of the build doesn't need it we can just ensure the same CFLAGS used for building are used in the probe macros.


kawasaki:
@stsquad I have merged this PR. Thanks!

PR #167, opened 2025-04-22
nvme/054: update to support all fabrics

yizhanglinux:


yizhanglinux:
@guixinliu1995

kawasaki:
@yizhanglinux The commit had an extra space, I removed it and then committed to the master branch as a69301b61f98. Thanks!

PR #169, opened 2025-05-09
[draft]add new case for NVMe namespace management

yizhanglinux:


tests/nvme/064
```diff
@@ -0,0 +1,65 @@
...
+	echo "Running ${TEST_NAME}"
+
+	local ns_dev ctrl_dev nn tnvmcap lbaf lbads cntlid
+
+	ns_dev=${TEST_DEV##*/}
+        ctrl_dev=${ns_dev%n*}
+	nn=$(nvme id-ctrl /dev/"$ctrl_dev" | grep ^nn | awk '{ print $3}')
+	tnvmcap=$(nvme id-ctrl /dev/"$ctrl_dev" | grep tnvmcap | awk '{ print $3}')
+	lbaf=$(nvme id-ns "$TEST_DEV" | grep 'in use' | awk '{ print $2}')
+	lbads=$(nvme id-ns -H "$TEST_DEV" | grep 'in use' | awk '{ print $12}')
```
igaw, line 33: Please use `--verbose` instead of the old `--human-readable`.

igaw:
Generally I would prefer if we would start using the JSON output instead the STDOUT of nvme-cli. The JSON output is there for machine parsing. Screen scraping is not ideal.

tests/nvme/064
yizhanglinux: Yeah, to check the Data Size, the --verbose cannot output that value of 512[1], it only output "ds":9[2], not sure if it's something we can improve on nvme-cli.
[1]
`LBA Format  1 : Metadata Size: 8   bytes - Data Size: 512 bytes - Relative Performance: 0x3 Degraded (in use)`
[2]
```
 "lbafs":[
    {
      "ms":0,
      "ds":9,
      "rp":1
    },
    {
      "ms":8,
      "ds":9,
      "rp":3
    },
    {
      "ms":0,
      "ds":12,
      "rp":0
    },
    {
      "ms":8,
      "ds":12,
      "rp":2
    },
    {
      "ms":64,
      "ds":12,
      "rp":3
    }
  ],

```

yizhanglinux:
> Generally I would prefer if we would start using the JSON output instead the STDOUT of nvme-cli. The JSON output is there for machine parsing.
> 
> This would mean we add dependency on something like `jq` but IMO this would be better than continuing with the screen scraping.

Thanks, just update to get the value from json ouput.

kawasaki:
Two comments:

1) I wanted to confirm the motivation to add this test case . Does it aim to test the kernel code? Or nvme devices? If it aim to cover kernel code, which part of the kernel code, or which feature of the kernel to cover?  (Namespace detection feature?)

2) Can we ensure that the test target device status does not change before and after the test case run? For example, I wonder what will happen when multiple namespaces belong to a single controller, and one of the namespaces is the test target device.

tests/nvme/064
igaw: Definitely something we can improve in nvme-cli. I understand that the verbose json output is lacking some information which is shown with stdout and verbose? Anyway, would you mind to open a bug report in nvme-cli so we don't forget about it. Thanks!

yizhanglinux: yes, I've filed below issue to track it.
https://github.com/linux-nvme/nvme-cli/issues/2811

PR #171, opened 2025-05-21
Multiple minor improvements for common scripts

bvanassche:


kawasaki:
Sorry for the late response. I misunderstood this PR would be relevant to the errexit discussion and delayed to look inside. But my mere understanding was wrong! Sorry for it. This is another improvement opportunity. Will look into this today.

common/fio
```diff
@@ -220,20 +220,23 @@ _run_fio_verify_io() {
...
-	local name field value
 	for name in "${FIO_PERF_FIELDS[@]}"; do
 		field="${FIO_TERSE_FIELDS["$name"]}"
 		if [[ -z $field ]]; then
 			echo "_fio_perf: unknown fio terse field '$name'" >&2
 			continue
 		fi
-		value="$(cut -d ';' -f "$field" "$TMPDIR/fio_perf")"
-		TEST_RUN["$FIO_PERF_PREFIX$name"]="$value"
+		TEST_RUN["$FIO_PERF_PREFIX$name"]=${value_array[$field]}
```
kawasaki, line 240: The right hand has array index off-by-one bug. It should be "${value_array[$((field - 1))]}".

kawasaki (changes requested):
Thanks for this PR. Overall, I think the changes are good. Please find my comments. I would like to have two fixes before merge.

common/fio
bvanassche: Thanks for the feedback. This issue has been fixed.

kawasaki:
Thank you for revising the commits! I have merged this PR.

PR #178, opened 2025-05-30
nvme: Add test for sending admin commands over disabled paths

hreinecke:
ANA states only apply to commands to individual namespaces, not to commands for the controller. So add a testcase to check if we can send admin commands over inaccessible paths.

### DIFF
--- a/tests/nvme/064
+++ b/tests/nvme/064
@@ -1,0 +1,53 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2024 Hannes Reinecke (SUSE) <hare@suse.de>
+#
+# Test nvme fabrics controller ANA failover during I/O
+
+. tests/nvme/rc
+
+DESCRIPTION="test admin commands on disabled paths"
+
+requires() {
+	_nvme_requires
+	_have_loop
+	_have_fio
+	_require_nvme_trtype_is_fabrics
+}
+
+set_conditions() {
+	_set_nvme_trtype "$@"
+}
+
+test() {
+	local -a ports
+	local port
+	local ns
+
+	echo "Running ${TEST_NAME}"
+
+	_setup_nvmet
+
+	_nvmet_target_setup
+
+	_nvme_connect_subsys
+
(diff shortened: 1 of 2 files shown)


tests/nvme/064
```diff
@@ -0,0 +1,53 @@
...
+	_setup_nvmet
+
+	_nvmet_target_setup
+
+	_nvme_connect_subsys
+
+	ns=$(_find_nvme_ns "${def_subsys_uuid}")
+	[ -z "${ns}" ] && return 1
+	nvme nvm-id-ctrl "/dev/${ns}"
+	nvme nvm-id-ns "/dev/${ns}"
```
igaw, line 38: Everytime `nvm-id-ctrl` or `nvm-id-ns` is touched this test will break. I think it would be better to test for expected values.

igaw: To make this test a bit future proof, it really should check for specific fields, as we do in nvme/059:

```shell
 nvme_awupf=$(nvme id-ctrl /dev/"${ctrl_dev}" | grep awupf | awk '{ print $3}')
```

BTW, I still think we should start using the json output for nvme-cli together with `jq`, e.g:

```shell
nvme id-ctrl /dev/nvme0 --output json | jq .awupf
```

kawasaki:
Today, I found a time slot to take a look in this PR. I renumbered the test case and ran it with the recent kernel and nvme-cli, I noticed that nvm-id-ctrl and nvm-id-ns commands just failed with inaccesible ANA , and printed the error messages below:

```
nvm identify controller: Resource temporarily unavailable
get-namespace-id: Resource temporarily unavailable
```

So it does not look meaningful to check output of these commands.

How about to redirect the command outputs to FULL file for logging, and just check the exit status of the nvme commands? Based on this idea, I modified the test script ([link](https://github.com/kawasaki/blktests/commit/306360359929d6da82c5478bd7ab97428ce1047c)).

kawasaki:
> In principle, yes. But the patch fixing that had been ... well, not directly ignored, but also not applied. On the grounds that there is no use-case. Maybe we can bring it up at LSF (at the blktest session?) to see how we can or should progress here.

Okay, sounds like a plan. Let's revisit this after the talk.

PR #180, opened 2025-06-09
loop/011: use lazy umount for tmpfs

yizhanglinux:
CKI reported random loop/011 failure, fix it by using lazy umount for tmpfs.

$ ./check loop/011
loop/011 (Make sure unsupported backing file fallocate does not fill dmesg with errors) loop/011 (Make sure unsupported backing file fallocate does not fill dmesg with errors) [failed]
    runtime  0.148s  ...  0.145sults/tmpdir.loop.011.QiG/tmpfs': Device or resource busy
    --- tests/loop/011.out	2025-06-09 03:49:23.017287762 +0000
    +++ /root/blk/results/nodev/loop/011.out.bad	2025-06-09 04:26:22.645147858 +0000
    @@ -1,3 +1,4 @@
     Running loop/011
    +umount: /root/blk/results/tmpdir.loop.011.QiG/tmpfs: target is busy.
     Found 1 error(s) in dmesg
     Test complete
$ mount | grep loop
testfs on /root/blk/results/tmpdir.loop.011.t7M/tmpfs type tmpfs (rw,relatime,seclabel,inode64)

### DIFF
--- a/tests/loop/011
+++ b/tests/loop/011
@@ -33,7 +33,7 @@ test() {
 	errors=$(_dmesg_since_test_start | grep -c "operation not supported error, dev .*WRITE_ZEROES")
 
 	losetup --detach "$loop_dev"
-	umount "$TMPDIR/tmpfs"
+	umount --lazy "$TMPDIR/tmpfs"
 
 	echo "Found $errors error(s) in dmesg"
 


tests/loop/011
```diff
@@ -33,7 +33,7 @@ test() {
 	errors=$(_dmesg_since_test_start | grep -c "operation not supported error, dev .*WRITE_ZEROES")
 
 	losetup --detach "$loop_dev"
-	umount "$TMPDIR/tmpfs"
+	umount -l "$TMPDIR/tmpfs"
```
kawasaki, line 36: It is preferred to have long option names. I suggest to use "--lazy" instead of "-l".

kawasaki:
The test case loop/011 aims to confirm the kernel fix 5f75e081ab5c. When the kernel has the fix commit, the test case passes. When the kernel does not have the commit, the test case fails. Using the kernel v6.13, I confirmed that this patch does not affect the capability to check the kernel fix. Looks good.

As I commented on the modified line, it is preferred to use the long options "--lazy" instead of the short option "-l". @yizhanglinux if you have time to afford, update the commit and this PR for the change? Or if you like, I can fold in the change and apply the commit. Please let me know your preference.

tests/loop/011
yizhanglinux: Thanks for the review, just updated.

PR #195, opened 2025-08-11
scsi/010: Add an F2FS data lifetime support test

bvanassche:
Test whether write hint information is passed to block devices.

### DIFF
--- a/tests/scsi/010
+++ b/tests/scsi/010
@@ -1,0 +1,71 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2025 Google LLC
+
+. tests/scsi/rc
+. common/null_blk
+. common/scsi_debug
+
+DESCRIPTION="test data lifetime propagation"
+QUICK=1
+
+requires() {
+	_have_fio
+	_have_driver f2fs
+	# See also commit 7643f3fe2772 ("f2fs: assign the write hint per stream
+	# by default"; v6.10).
+	_have_kver 6 10
+	_have_program mkfs.f2fs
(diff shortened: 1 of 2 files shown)


tests/scsi/010
```diff
@@ -0,0 +1,63 @@
...
+	_have_driver f2fs
+	# See also commit 7643f3fe2772 ("f2fs: assign the write hint per stream
+	# by default"; v6.10).
+	_have_kver 6 10
+	_have_program mkfs.f2fs
+	_have_scsi_debug_group_number_stats
+}
+
+test() {
+	set -e
```
kawasaki, line 23: This test case will be the second test case to use set -e (errexit) feature. When an error happens between the set -e and set +e, scsi_debug device could be left in mounted status, and could affect to the following test cases. I suggest to factor out the region between set -e and set +e as a function, as you did for zbd/013.

tests/scsi/010
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2025 Google LLC
+
+. tests/scsi/rc
+. common/null_blk
+. common/scsi_debug
+
+DESCRIPTION="test F2FS data lifetime support"
```
kawasaki, line 9: This description and the commit message title both confused me (why not fstests?). So I suggest "test write hint  integration with F2FS data lifetime".

kawasaki (changes requested):
Other than the 2 comments I made, this test case looks good to me. Thanks.

tests/scsi/010
bvanassche: This has been fixed.

PR #198, opened 2025-08-21
common/rc: add _have_crypto()

hreinecke:
Add a function _have_crypto() to check if a specific crypto algorithm is present. Some cryptographic function are implemented directly in the kernel, so a simple '_have_driver' doesn't work here.

### DIFF
--- a/common/rc
+++ b/common/rc
@@ -68,6 +68,30 @@ _have_driver()
 	return 0
 }
 
+# Check that the specified crypto algorithm is present, regardless of
+# whether it's build-in, a module, or provided by the kernel.
+# The module will be loaded if it is loadable, and unloaded at test case end.
+_have_crypto()
+{
+	local modname="${1//-/_}"
+
+	if [ -d "/sys/module/${modname}" ]; then
+		return 0
+	fi
+
+	if ! modprobe -q "${modname}"; then
+		if ! grep -q "${modname}" /proc/crypto; then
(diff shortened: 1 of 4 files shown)


igaw:
Shouldn't the crypto algorithm be just `DH`? The output from `/proc/crypto` says:

```
name         : dh
driver       : dh-generic
module       : kernel
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : kpp
```

This works with the driver name, just wondering how to make it future proof. Is the driver name stable? Or would it better to match on the name? Obviously, then the `modprobe` is not straight forward. 

Anyway, I this should address my problem so I am fine with it as it is. :)

common/rc
```diff
@@ -68,6 +68,30 @@ _have_driver()
 	return 0
 }
 
+# Check that the specified crypto algorithm is present, regardless of
+# whether it's build-in, a module, or provided by the kernel.
```
bvanassche, line 72: build-in -> built-in

kawasaki:
Let me confirm. I understand that this will avoid the confusion found in @igaw system. How about "_have_kernel_option CRYPTO_DH" instead of this commit? If it works, I think it's cleaner. If it doesn't work, I'm okay with this fix approach.

As to the change, the commit introduces some code duplication between _have_driver() and _have_crypto(). It is the better to avoid it. Maybe I can do a follow-up commit afterwards.

PR #214, opened 2025-11-19
nvme/050: add support for NVMe multipath devices

yizhanglinux:
$./check nvme/050
nvme/050 => nvme1n1 (test nvme-pci timeout with fio jobs)    [failed]
    runtime  94.236s  ...  62.734s
    --- tests/nvme/050.out	2025-11-17 00:23:56.086469327 -0500
    +++ /root/blktests/results/nvme1n1/nvme/050.out.bad	2025-11-19 03:17:45.389644408 -0500
    @@ -1,2 +1,3 @@
     Running nvme/050
    -Test complete
    +Test failed
    +tests/nvme/050: line 50: /sys/bus/pci/devices//remove: Permission denied

### DIFF
--- a/common/multipath-over-rdma
+++ b/common/multipath-over-rdma
@@ -156,7 +156,7 @@ dev_to_mpath() {
 	[ -e "$1" ] || return $?
 
 	if [ -L "$1" ]; then
-		e=$(readlink -f "$1")
+		e=$(readlink --canonicalize "$1")
 	else
 		e="$1"
 	fi
@@ -167,7 +167,7 @@ dev_to_mpath() {
 
 	for d in /dev/mapper/mpath*; do
 		if [ -L "$d" ]; then
-			e=$(readlink -f "$d")
+			e=$(readlink --canonicalize "$d")
 		elif [ -e "$d" ]; then
 			e="$d"
 		else
@@ -240,7 +240,7 @@ remove_mpath_dev() {
 				break
 			fi
 			ls -l "$1"
-			stop_bdev_users "$(readlink -f "$1")"
+			stop_bdev_users "$(readlink --canonicalize "$1")"
 			sleep .5
 		done
 		if [[ $i = 0 ]]; then
--- a/common/nvme
+++ b/common/nvme
@@ -1122,7 +1122,7 @@ _nvmet_target_cleanup() {
 }
 
 _require_test_dev_is_nvme() {
-	if ! readlink -f "$TEST_DEV_SYSFS/device" | grep -q nvme; then
+	if ! readlink --canonicalize "$TEST_DEV_SYSFS/device" | grep -q nvme; then
 		SKIP_REASONS+=("$TEST_DEV is not a NVMe device")
 		return 1
 	fi
--- a/common/rc
+++ b/common/rc
@@ -392,16 +392,16 @@ _test_dev_set_scheduler() {
 }
 
 _require_test_dev_is_pci() {
-	if ! readlink -f "$TEST_DEV_SYSFS/device" | grep -q pci; then
+	if ! readlink --canonicalize "$TEST_DEV_SYSFS/device" | grep -q pci; then
 		# nvme needs some special casing
-		if readlink -f "$TEST_DEV_SYSFS/device" | grep -q nvme; then
+		if readlink --canonicalize "$TEST_DEV_SYSFS/device" | grep -q nvme; then
 			local blkdev
 			local ctrldev
 			# First get the controller device from the namespace blockdev
 			blkdev="$(echo "$TEST_DEV_SYSFS" | cut -d '/' -f 7)"
 			ctrldev="$(echo "$blkdev" | grep -Eo 'nvme[0-9]+')"
 			# Then get the pci device from the controller device
-			if readlink -f "$ctrldev/device" | grep -1 pci; then
+			if readlink --canonicalize "$ctrldev/device" | grep -1 pci; then
 				return 0
 			fi
 		fi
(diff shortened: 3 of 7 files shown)


igaw:
There are enterprise PCI disks with multipath. I think it would be good to tests these devices as well. It seems you are testing with such a device, maybe we could get this tests also working with these types?

I think it it's possible to make `_get_pci_dev_from_blkdev` a bit smarter so it returns the correct PCI device?

yizhanglinux:
@kawasaki This can skip the test when the disk is an enterprise PCI disk with multipath.
I think what @igaw suggests is we still need to test such enterprise pci disk with multiapth.

Maybe something like below to return the PCI device when the disk supports multipath:

```
diff --git a/tests/nvme/050 b/tests/nvme/050
index 91f3564..ba09956 100755
--- a/tests/nvme/050
+++ b/tests/nvme/050
@@ -25,8 +25,7 @@ test_device() {
        local i

        echo "Running ${TEST_NAME}"
-
-       pdev=$(_get_pci_dev_from_blkdev)
+       pdev=$(_nvme_get_pci_from_dev_sysfs)
        nvme_ns="$(basename "${TEST_DEV}")"
        echo 1 > /sys/block/"${nvme_ns}"/io-timeout-fail

diff --git a/tests/nvme/rc b/tests/nvme/rc
index a8f80d8..e25cda2 100644
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -87,6 +87,24 @@ _require_test_dev_is_not_nvme_multipath() {
        return 0
 }

+_nvme_dev_support_native_multipath() {
+       if [[ "$(readlink -f "$TEST_DEV_SYSFS/device")" =~ /nvme-subsystem/ ]]; then
+               return 0
+       fi
+       return 1
+}
+
+_nvme_get_pci_from_dev_sysfs() {
+       if _nvme_dev_support_native_multipath; then
+               readlink -f /sys/block/$(basename "${TEST_DEV}")/multipath/nvme*c*n*/device | \
+                       grep -Eo '[0-9a-f]{4,5}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]' | \
+                       tail -1
+       else
+               readlink -f "$TEST_DEV_SYSFS/device" | \
+                       grep -Eo '[0-9a-f]{4,5}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]' | \
+                       tail -1
+       fi
+}
 _require_test_dev_support_sed() {
        if ! nvme sed discover "$TEST_DEV" &> /dev/null; then
                SKIP_REASONS+=("$TEST_DEV doesn't support SED operations")
```

kawasaki:
@yizhanglinux Thanks. Overall, the suggested change looks good. Posting it as a proper patch or PR will be appreciated. Again, please use "--canonicalize" instead of "-f". Also, I think it is good to have another patch to replace "-f" with "--canonicalize" for other places, which can be done in the same PR/series, or later.

yizhanglinux:
> 
> or if we don't want to trust that when a cmic bit is set it's multipath device then we should check what the sysfs is telling us.

Check cmic maybe not enough, we also need to check the nvme_core.multipath=Y.

Maybe it's better to just check the sysfs:
```
diff --git a/tests/nvme/050 b/tests/nvme/050
index 91f3564..2fdf62e 100755
--- a/tests/nvme/050
+++ b/tests/nvme/050
@@ -20,13 +20,11 @@ requires() {
 }

 test_device() {
-       local nvme_ns
-       local pdev
-       local i
+       local nvme_ns pdev io_error i

        echo "Running ${TEST_NAME}"

-       pdev=$(_get_pci_dev_from_blkdev)
+       pdev=$(_nvme_get_pci_from_dev_sysfs)
        nvme_ns="$(basename "${TEST_DEV}")"
        echo 1 > /sys/block/"${nvme_ns}"/io-timeout-fail

@@ -40,10 +38,13 @@ test_device() {
            --name=reads --direct=1 --filename="${TEST_DEV}" --group_reporting \
            --time_based --runtime=1m >& "$FULL"

-       if grep -q "Input/output error" "$FULL"; then
-               echo "Test complete"
+       io_error=false
+       grep -q "Input/output error" "$FULL" && io_error=true
+
+       if _nvme_dev_support_native_multipath; then
+               $io_error && echo "Test failed" || echo "Test complete"
        else
-               echo "Test failed"
+               $io_error && echo "Test complete" || echo "Test failed"
        fi

        # Remove and rescan the NVME device to ensure that it has come back
diff --git a/tests/nvme/rc b/tests/nvme/rc
index a8f80d8..2ecf618 100644
--- a/tests/nvme/rc
+++ b/tests/nvme/rc
@@ -87,6 +87,27 @@ _require_test_dev_is_not_nvme_multipath() {
        return 0
 }

+_nvme_dev_support_native_multipath() {
+       if [[ "$(readlink -f "$TEST_DEV_SYSFS/device")" =~ /nvme-subsystem/ ]]; then
+               return 0
+       fi
+       return 1
+}
+
+_nvme_get_pci_from_dev_sysfs() {
+       if _nvme_dev_support_native_multipath; then
+               readlink -f /sys/block/$(basename "${TEST_DEV}")/multipath/nvme*c*n*/device | \
+                       grep -Eo '[0-9a-f]{4,5}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]' | \
+                       tail -1
+       else
+              _get_pci_dev_from_blkdev
+       fi
+}
+
 _require_test_dev_support_sed() {
        if ! nvme sed discover "$TEST_DEV" &> /dev/null; then
                SKIP_REASONS+=("$TEST_DEV doesn't support SED operations")
```

igaw:
Yeah, the default operation is " echo 1 > /sys/block/nvme0n1/io-timeout-fail", do you mean also add echo 1 > /sys/block/nvme0c0n1/io-timeout-fail here?

Yes, the failure injection happens at this level, that means you have to set it for all nvme0c*n1 devices. Another thing you need to do then is to set the default timeout for the I/Os. The default value is 30s. So the block layer will hold all I/Os in the queue and they won't be reported to the upper layers before it times out.

Something like this:

    set fail_io_timeout for all nvme0c*n1 devices
    set /sys/block/nvme0n1/queue/timeout relative short, let's say 1 second
    let fio run
    add sleep 2, give the block layer a chance to return an error.
    check for errors, it should fail now

PR #215, opened 2025-11-25
common/rc: add _have_crypto_algorithm

igaw:
A crypto algorithm might be provided as built-in, thus check first if it is always present before trying to load as module.

The idea to check the CONFIG flags is not enough. The algorithm needs to be present in the kernel. `/proc/crypto` has all the info. `/sys/modules/*` will not show built-in modules.

Fixes: #https://github.com/linux-blktests/blktests/issues/197
Replaces: #https://github.com/linux-blktests/blktests/pull/198

### DIFF
--- a/common/rc
+++ b/common/rc
@@ -69,6 +69,19 @@ _have_driver()
 	return 0
 }
 
+# Check that the specified crypto algorithm is present, regardless of whether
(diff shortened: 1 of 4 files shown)


igaw:
Your changes look good to me. Thanks!

PR #216, opened 2025-11-26
check `nvme connect` return codes

igaw:
Increase test coverage by checking if `nvme connect` is also reporting the expected return code.

At least nvme/031 seems to trigger sometimes FAILs.

To my surprise, nvme/041 for FC is passing with https://lore.kernel.org/linux-nvme/20251117184343.97605-1-justintee8345@gmail.com/  applied. Nice :)


igaw:
Thanks for applying them.

PR #221, opened 2025-11-28
avoid test timeouts caused by scsi_debug delay

gitPo1son:


kawasaki:
I ran the throtl test group using the old kernel v6.17-rc3 and blktests with this PR. Then I confirmed that the test case throtl/002 failed with the symptom that recorded in the commit d91fa932fbaa. It indicates that the testing capability of throtle group is not affected by this PR. Good.

kawasaki:
@gitPo1son Thanks for improving the commit message. I did a bit more improvement in the commit message, and applied it to the master branch. This PR was not merged, but its purpose has been achieved. Thank you :) Let me close this.

PR #228, opened 2026-03-13
Add nvme subsystem reset test

yizhanglinux:


tests/nvme/rc
```diff
@@ -99,6 +99,14 @@ _require_test_dev_support_sed() {
 	return 1
 }
 
+_require_test_dev_support_subsystem_reset() {
+	if ! nvme show-regs "$TEST_DEV" -H | grep -q "NSSRS.*Yes"; then
```
igaw: FWIW, `show-regs` is very likely almost never going to work.

PR #229, opened 2026-03-13
nvme/068: add new test for nvme subsystem-reset test

yizhanglinux:


tests/nvme/068
```diff
@@ -0,0 +1,49 @@
...
+
+	ctrl_dev=${TEST_DEV%n*}
+
+	# Start nvme subsystem-reset operation
+	if ! nvme subsystem-reset "$ctrl_dev" >> "$FULL" 2>&1; then
+                echo "ERROR: subsystem-reset failed"
+        fi
+
+	# Wait NVMe disk reinitialized
+	sleep 10
```
shroffni: At least on PPC platform, adding delay of sleep 10 would not work. As we know, running subsystem-reset command on PPC platform would cause the communication loss to the NVMe adapter. So on PPC system any I/Os those were running while susbsytem-reset is executed or any new I/O submitted after the subsystem-reset is executed would eventually times out after 30 seconds. The nvme timeout handler code would then attempt to read PCIe/MMIO config space register which triggers the EEH and then EEH would recover the communication link to the NVMe adapter. So in theory, in worst case, it would take more than 30 seconds for the link to be restored and device to be back online post subsystem-reset. 

I'd suggest following steps:
1. Start I/O (maybe using fio)
2. execute nvme subsystem-reset 
3. Add sleep 35 (30 seconds for I/O request timeout plus additional 5 seconds as cushion)

PR #240, opened 2026-03-31
loop: add regression test for partscan double-scan race

daandemeyer:
Add a stress test that detects spurious partition removal events when setting up a loop device with partscan enabled.

The kernel bug was that disk_force_media_change() set GD_NEED_PART_SCAN, causing udev's device open to trigger a partition scan racing with the explicit scan from loop_reread_partitions(). The second scan would drop and re-add all partitions, making partition devices briefly disappear.

The test monitors kernel uevents while repeatedly setting up and tearing down a loop device with partscan. Each cycle should produce exactly one add and one remove uevent for the partition device. Extra events indicate the double-scan race was triggered.

Link: https://lore.kernel.org/linux-block/20260330081819.652890-1-daan@amutable.com/T/#u

### DIFF
--- a/common/rc
+++ b/common/rc
@@ -603,6 +603,22 @@ _systemd_start_udevd() {
 	return 0
 }
 
+_have_systemd_ver() {
+	local required_ver=${1}
+	local ver
+
+	if ! command -v systemctl &>/dev/null; then
+		SKIP_REASONS+=("systemd is not available")
+		return 1
+	fi
+
+	ver=$(systemctl --version | head -1 | sed 's/systemd //;s/[^0-9].*//')
+	if ((ver < required_ver)); then
+		SKIP_REASONS+=("systemd version is older than ${required_ver}")
+		return 1
+	fi
+}
+
 # Run the given command as NORMAL_USER
 _run_user() {
 	su "$NORMAL_USER" -c "$1"
--- a/tests/loop/012
+++ b/tests/loop/012
@@ -1,0 +1,77 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 Daan De Meyer
+#
+# Regression test for a race between udev and loop_reread_partitions().
+#
+# When LOOP_CONFIGURE is called with LO_FLAGS_PARTSCAN,
+# disk_force_media_change() used to set GD_NEED_PART_SCAN before the
+# uevent was sent. When udev opened the device in response,
+# blkdev_get_whole() would trigger a partition scan, and then
+# loop_reread_partitions() would scan again. The second scan drops all
+# partitions from the first scan before re-adding them, causing
+# partition devices to briefly disappear.
+#
+# Verify that setting up a loop device with partscan does not produce
+# spurious partition add/remove events.
+
+. tests/loop/rc
+
+DESCRIPTION="check for spurious partition removal when partscan is enabled"
+TIMED=1
+
+requires() {
+	_have_program sfdisk
+	_have_systemd_ver 259
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+
+	truncate -s 3MiB "$TMPDIR/img"
+	sfdisk "$TMPDIR/img" >"$FULL" 2>&1 <<-EOF
+		label: gpt
+		size=1MiB
+	EOF
+
+	local dev
+	dev="$(losetup -f)"
+
+	# Monitor kernel uevents for partition block devices.
+	udevadm monitor --kernel -s block/partition >"$TMPDIR/uevents" 2>&1 &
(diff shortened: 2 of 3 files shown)


kawasaki:
Hello @daandemeyer

Today, I ran the added test case on my test environment, and found the test case fail even when I apply the fix patch to the kernel.

```
$ sudo ./check loop/012
loop/012 (check for spurious partition removal when partscan is enabled) [failed]
    runtime  5.682s  ...  6.243s
    --- tests/loop/012.out      2026-04-08 10:11:48.266000000 +0900
    +++ /home/shin/Blktests/blktests/results/nodev/loop/012.out.bad     2026-04-08 13:45:01.585000000 +0900
    @@ -1,2 +1,4 @@
     Running loop/012
    +Fail: 111 iterations but 149 add events (expected 111)
    +Fail: 111 iterations but 149 remove events (expected 111)
     Test complete
[shin@testnode2 blktests ]$
```

I observed the failure with both v2 and v3 kernel patches. I used v7.0-rc7 as the baseline kernel. Is this failure expected?

kawasaki:
One more nit comment. Some commands in the new test case use short options: "truncate -s", "losetup -f" or so. For readability and stability, longer options are recommended, like "truncate --size=", "losetup --find". I'm fine with the current short options, but if you have chance to respin, I suggest to use longer options. I attach a diff file to show how the new test case will look like with the longer options.

[review.txt](https://github.com/user-attachments/files/26558635/review.txt)

daandemeyer:
@kawasaki I ran it a bunch of times and it doesn't fail on my test setup :/

kawasaki:
@daandemeyer Thank you for trying out. FYI,  here I attach the kernel config of the kernel that I used.
[_config_v7.0-rc7_pr240_failure.gz](https://github.com/user-attachments/files/26669427/_config_v7.0-rc7_pr240_failure.gz)

And I used Fedora 43 on QEMU VM as the userland of my test environment. Which distro do you use? Do you run your test on any VM or bare metal machine?

daandemeyer:
My test environment is https://github.com/DaanDeMeyer/mkosi-kernel, an Arch Linux image running in qemu (booted with systemd).

(I verified the test failed without my changes as well)

kawasaki:
Thanks. I started trying mkosi and mkosi-kernel, but still failing to building the Arch image. Will try to allocate more time for it.

kawasaki:
I managed to run mkosi with mkosi-kernel to generate Fedora and Arch images. (I fell in the pit fall that Fedora 43 kernel failed to boot with the Fedora rawhide image generated by mkosi. When I specify kernel source to build, the failure disappeared.)

With the Fedora and Arch images, I confirmed that the new test case works as expected. The test case fails without the kernel side fix, and the test case passes with the kernel side fix.

Then, the next step is to investigate why the test case fails on my Fedora QEMU environment even with the kernel side fix. Now I'm chasing this.

kawasaki:
I set up "Distribution=fedora" in mkosi.local.conf. I tried both "Release=rawhide" and "Release=43", and found that the test case passes with "Release=rawhide", but fail with "Release=43" (same kernel v7.0 + kernel fix by @daandemeyer ). I guess systemd version difference between the Fedora version might be the cause of the behavior difference.
-  Fedora rawhide: systemd 260.1
- Fedora 43: systemd 258.7

If this hypothesis is correct, it means that the new test case has systemd version dependency. I would like to confirm it, but I do not know how to replace systemd. @daandemeyer If you have any good idea to confirm it, please share. Can mkosi replace systemd version?

kawasaki:
If was stupid to ask "Can mkosi replace systemd version?". mkosi is the systemd origin side project to build up the OS image to help systemd development. Of course it can. And @daandemeyer already documented how to do it in README of mkosi-kernel. It was a bit tough for newbees, but I have managed to control both systemd and kernel versions to build Fedora rawhide image.

With that environment, I confirmed that the new test case depends on both kernel and systemd:

- The kernel fix is required to make the test case pass
- Even with the kernel fix, systemd version 259 is required to make the test case pass

I bisected further, and found that the systemd commit [aa47d8ade18c](https://github.com/poettering/systemd/commit/aa47d8ade18cc4a079fef5a1aaa37d763507104e) is required to make the test case pass. The test case fails with systemd v258 tag or before the commit.

Based on this observation, I suggest to check systemd version in the required() function. If it is older than v259, skip the test case. I attach a patch file [patch.txt](https://github.com/user-attachments/files/26969699/patch.txt) for reference. @daandemeyer Please let me know your thoughts.

daandemeyer:
@kawasaki Thank you for the investigation! The attached patch looks good to me. I will apply it and update the PR

kawasaki:
Thank you for the update. I have merged this PR.

PR #243, opened 2026-04-08
tests: update to use _get_page_size to get PAGE_SIZE

yizhanglinux:


kawasaki:
@yizhanglinux Thanks! I had noticed that this change is beneficial when you introduced _get_page_size().

As to the patch, I would like to ask three things: 1) zbd/014 has "_have_program getconf" in requires(). This line should be remove together in the patch. 2) The commit message is empty and its title is rather unclear. Let's add test cases modified in the commit title, and explain why this change is good. I suggest the commit message as follows.

```
    throtl/{002,003,007}, zbd/014: use _get_page_size to get PAGE_SIZE

    Commit 8eca9fa369ee ("common/rc, scsi/011, zbd/010: introduce
    _page_size_equals() helper") introduced the helper bash function
    _get_page_size(). Replace "getconf PAGE_SIZE" with the helper to cut
    dependency to the getconf tool.
```

Also, could you post this patch to linux-block list? This change is not a big deal, but I would like to make sure the change was shared on the list before apply. If you are too busy to handle this, please let me know so that I can help it out.

kawasaki:
Per local talk, I modified the patch and [posted it to linux-block](https://lore.kernel.org/linux-block/20260416114235.4076679-1-shinichiro.kawasaki@wdc.com/) on behalf of @yizhanglinux.

kawasaki:
I applied the modified change. Thank goes to @yizhanglinux. I close this case.

PR #245, opened 2026-06-11
nvme/059: fix unset bytes_to_write in TEST 7

schlad:
bytes_to_write was never assigned, causing TEST 7 to pass for the wrong reason. Set it to atomic_unit_max_bytes + logical_block_size and update the golden output with the expected "pwrite: Invalid argument" from xfs_io.


kawasaki:
@schlad Good catch, thanks! The commit looks good to me.

Actually, I found similar mistake exists in scsi/009 TEST 8 also. If you can create another PR or post patch for scsi/009, it will be a big help. (If you do not have time for it, that's okay, just let me know.)

@alanpeterad FYI. If you have any comment, please share.

schlad:
> If you can create another PR or post patch for scsi/009, it will be a big help.

I will take a look at the scsi/009 today afternoon (European time zone here).

PR #248, opened 2026-07-13
Handle nvme cli 3.x output changes

### LINKED ISSUE #247: how to handle nvme-cli output behavior change
I've merged a bigger behavior change into nvme-cli which makes the tool
behave more like other Unix tools.  Emitting status messages by
default violates the long-established convention that stdout is reserved
for machine-consumable data, while diagnostics and optional progress
reporting belong on stderr or behind an explicit verbose flag

See here for more details:  https://github.com/linux-nvme/nvme-cli/pull/3553

As result all fabric related blktests will fail:

https://github.com/linux-nvme/nvme-cli/actions/runs/28856830090/job/85585355343
```
[93](https://github.com/linux-nvme/nvme-cli/actions/runs/28856830090/job/85585355343#step:3:2497)
    +++ /home/runner/blktests/results/nodev/md/001.out.bad	2026-07-07 09:52:26.470476655 +0000
    @@ -1,3 +1,2 @@
     Running md/001
    -disconnected 1 controller(s)
     Test complete
```

It's possible to re-enable those outputs by using the `--verbose` flag (though I have to double check
if the exact same output is generated).

My question is how do want to handle this? My preferences here would be to use only
the status code to decided if the test fail or passes. We could also try to figure out which
nvme-cli version is used and adapt the tests accordingly. WDYT?

dwsuse:
nvme-cli 3.x changes its output behavior to match standard POSIX tools. This means there is no output on success, and status and error messages are sent to stderr. For nvme commands, it is sufficient to check the return code to determine whether the command was successful.

By relying solely on the return code, backward compatibility with nvme-cli 2.x is also ensured.

Fixes: #247

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -264,23 +264,52 @@ _setup_nvmet() {
 
 _nvme_disconnect_ctrl() {
 	local ctrl="$1"
+	local subsysnqn="$def_subsysnqn"
+	local no_wait=false
 
-	nvme disconnect --device "${ctrl}"
-}
+	shift
 
-_nvme_disconnect_ctrl_sync() {
-	local ctrl="$1"
-	local subsysnqn="$2"
-
-	_nvme_disconnect_ctrl "${ctrl}"
-
-	for ((i = 0; i < 10; i++)); do
-		_nvme_ctrl_ready "${ctrl}" "${subsysnqn}"
-		if [[ $? -eq 1 ]]; then
-			break
-		fi
-		sleep .1
+	while [[ $# -gt 0 ]]; do
+		case $1 in
+			--subsysnqn)
+				subsysnqn="$2"
+				shift 2
+				;;
+			--no-wait)
+				no_wait=true
+				shift 1
+				;;
+			*)
+				echo "WARNING: unknown argument: $1"
+				shift
+				;;
+		esac
 	done
+
+	local ctrl_dir="/sys/class/nvme/${ctrl}"
+	if [[ -d "${ctrl_dir}" ]]; then
+		local ret
+
+		nvme disconnect --device "${ctrl}" &> "${FULL}"
+		ret=$?
+		if [[ ${ret} -ne 0 ]]; then
+			echo "FAIL: nvme disconnect returned error code: ${ret}"
+		fi
+
+		if [[ ${no_wait} = false ]]; then
+			for ((i = 0; i < 10; i++)); do
+				_nvme_ctrl_ready "${ctrl}" "${subsysnqn}"
+				if [[ $? -eq 1 ]]; then
+					break
+				fi
+				sleep .1
+			done
+		else
+			if [[ -d "${ctrl_dir}" ]]; then
+				echo "FAIL: Controller ${ctrl} still exists after disconnect"
+			fi
+		fi
+	fi
 }
 
 _nvme_wait_subsys_removed() {
@@ -291,13 +320,19 @@ _nvme_wait_subsys_removed() {
 		if [ "$subsysnqn" == "$_subsysnqn" ]; then
 			for ((i = 0; i < 10; i++)); do
 				if [ ! -d "$subsyspath" ]; then
-					break
+					return 0
 				fi
 				sleep .1
 			done
-			break
+
+			if [ -d "$subsyspath" ]; then
+				return 1
+			fi
+			return 0
 		fi
 	done
+
+	return 0
 }
 
 _nvme_connect_subsys() {
@@ -503,6 +538,7 @@ _nvme_connect_subsys() {
 
 _nvme_disconnect_subsys() {
 	local subsysnqn="$def_subsysnqn"
+	local ret
 
 	while [[ $# -gt 0 ]]; do
 		case $1 in
@@ -517,8 +553,15 @@ _nvme_disconnect_subsys() {
 		esac
 	done
 
-	nvme disconnect --nqn "${subsysnqn}" |& tee -a "$FULL" |
-		grep -o "disconnected.*"
+	nvme disconnect --nqn "${subsysnqn}" &> "${FULL}"
+	ret=$?
+	if [[ ${ret} -ne 0 ]]; then
+		echo "FAIL: nvme disconnect returned error code: ${ret}"
+	fi
+
+	if ! _nvme_wait_subsys_removed "${subsysnqn}"; then
+		echo "FAIL: subsystem sysfs entry still exits"
+	fi
 }
 
 _nvme_ctrl_ready() {
--- a/tests/nvme/014
+++ b/tests/nvme/014
@@ -28,6 +28,7 @@ test() {
 	local size
 	local bs
 	local count
+	local ret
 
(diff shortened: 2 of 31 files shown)


schlad:
Tested against openSuse Tumbleweed - as far as this test run is concerned - **all looks good**.

See for details:
before: https://openqa.opensuse.org/tests/6092955
after (@dwsuse branch): https://openqa.opensuse.org/tests/6094793

Also internally tested against `nvme-cli-2.16` - all seems to be good including transports  - `fc` and `tcp`.

kawasaki:
Thanks for this PR. This approach looks good to me. I will leave some minor comments in line.

tests/nvme/014
```diff
@@ -42,7 +42,11 @@ test() {
 	dd if=/dev/urandom of="/dev/${ns}" \
 		count="${count}" bs="${bs}" status=none
 
-	nvme flush "/dev/${ns}"
+	nvme flush "/dev/${ns}" > /dev/null 2>&1
```
kawasaki: How about to redirect to "$FULL" instead of /dev/null. If any failure happens, the full file will be helpful.
Also, '> X 2>&1' can be shortened into '&> X'. I'm okay with the current form, but you may like the shorter form.

tests/nvme/014
```diff
@@ -42,7 +42,11 @@ test() {
...
+	ret=$?
```
kawasaki, line 47: Nit: local var declaration for 'ret' looks missing.

tests/nvme/014
```diff
@@ -42,7 +42,11 @@ test() {
...
+	if [[ ${ret} -ne 0 ]]; then
+			echo "FAIL: nvme flush return error code: $(ret)"
```
kawasaki: Extra indent here?

tests/nvme/054
```diff
@@ -27,6 +27,33 @@ resv_report() {
 		grep -E "gen|rtype|regctl|regctlext|cntlid|rcsts|rkey"
 }
 
+resv_register() {
```
kawasaki: The three functions resv_register, resv_acquire and resv_release are almost same. I suggest to combine them into a function:
```
nvme_resv() {
       local cmd=$1
       local test_dev=$2
       shift 2
       if ! nvme resv-"$cmd" "$test_dev" "$@" >/dev/null 2>&1; then
               echo "FAIL: nvme resv-$cmd failed on $test_dev with args: $*"
                return 1
        fi
}
```
With this, "resv_register ..." can be replaced with "nvme_resv register ..."

tests/nvme/019
```diff
@@ -33,8 +33,14 @@ test() {
...
 
 	ns=$(_find_nvme_ns "${def_subsys_uuid}")
+
 	nvme dsm "/dev/${ns}" --ad \
-		--slbs "${sblk_range}" --blocks "${nblk_range}"
+			--slbs "${sblk_range}" \
+			--blocks "${nblk_range}" > /dev/null 2>&1
+	ret=$?
+	if [[ ${ret} -ne 0 ]]; then
+			echo "FAIL: nvme dsm return error code: $(ret)"
```
kawasaki: Extra indent?

tests/nvme/019
```diff
@@ -33,8 +33,14 @@ test() {
 	_nvme_connect_subsys
 
 	ns=$(_find_nvme_ns "${def_subsys_uuid}")
+
 	nvme dsm "/dev/${ns}" --ad \
-		--slbs "${sblk_range}" --blocks "${nblk_range}"
+			--slbs "${sblk_range}" \
+			--blocks "${nblk_range}" > /dev/null 2>&1
+	ret=$?
```
kawasaki, line 41: Nit: local var declaration for ret.

common/nvme
```diff
@@ -264,23 +264,51 @@ _setup_nvmet() {
...
+			*)
+				echo "WARNING: unknown argument: $1"
+				shift
+				;;
+		esac
+	done
 
-	_nvme_disconnect_ctrl "${ctrl}"
+	local ctrl_dir="/sys/class/nvme/${ctrl}"
+	if [[ -e "${ctrl_dir}" ]]; then
```
kawasaki: Later on, -d check is used instead of -e for ${ctrl_dir}. For consistency, I suggest to use -d here also.

common/nvme
```diff
@@ -264,23 +264,51 @@ _setup_nvmet() {
...
-		_nvme_ctrl_ready "${ctrl}" "${subsysnqn}"
-		if [[ $? -eq 1 ]]; then
-			break
+		if [[ ${no_wait} = false ]]; then
+			for ((i = 0; i < 10; i++)); do
+				_nvme_ctrl_ready "${ctrl}" "${subsysnqn}"
+				if [[ $? -eq 1 ]]; then
+					break
+				fi
+				leep .1
```
kawasaki: I guess you meant 'sleep', not 'leep'.

common/nvme
```diff
@@ -264,23 +264,51 @@ _setup_nvmet() {
...
+
+				done
```
kawasaki: Indent is broken here.

common/nvme
```diff
@@ -264,23 +264,51 @@ _setup_nvmet() {
...
+				shift
+				;;
+		esac
+	done
 
-	_nvme_disconnect_ctrl "${ctrl}"
+	local ctrl_dir="/sys/class/nvme/${ctrl}"
+	if [[ -e "${ctrl_dir}" ]]; then
+		nvme disconnect --device "${ctrl}" > /dev/null 2>&1
+		ret=$?
```
kawasaki, line 294: Nit: local var declaration for 'ret'.

kawasaki (changes requested):
Overall, the change looks good to me. Please check my comments and see if they make sense for you. Thanks!

schlad:
I gave it another go since there are some changes - no surprises -  as far as this test run is concerned, all looks good. Tested on openSuse with nvme-cli 3.0

kawasaki:
@dwsuse Thanks for the updates. The commits look good. I also did some test runs and found no issue. I think this PR is ready for merge. When do you plan to release nvme-cli 3.0? If it is not so far away, I think it's okay to merge this PR soon.

igaw:
There are still a few things I would like to go over before the release and just this week the new config file format has landed. Realistically speaking I'd say it's at least one month until the release. 

I've already started with beta releases and packaged it up for [Tumbleweed](https://build.opensuse.org/package/show/openSUSE:Factory/nvme-cli). That is our QA team is already testing nvme-cli 3 with blktests. Thus @schlad is eager to get this fix into blktests so our testing is green again

Also our nvme-cli nightly builds do fail due to blktests: https://github.com/linux-nvme/nvme-cli/actions/runs/30066255452/job/89397709944#step:3:2516

Or in other words: could we get this merged soon? Thanks a lot!

kawasaki:
@igaw Good to know the nvme-cli 3.0 release outlook. I've merged this PR. Thanks!

PR #250, opened 2026-07-15
loop: test cleanup of manually added partitions

daandemeyer:
Partitions can be added to a loop device with BLKPG while LO_FLAGS_PARTSCAN is clear. After disk_force_media_change() stopped setting GD_NEED_PART_SCAN, such partitions could survive both detach and reopening the unbound loop device.

Add a regression test that creates a partition with partx, detaches the loop device, reopens it, and verifies that the partition is gone. Stop an active systemd-udevd during the test so userspace cannot hide missing kernel cleanup, and restore it through the test cleanup hook.

Opening the device after detach allows both the historical lazy cleanup and the new eager teardown cleanup to pass while detecting the regression between them.

Link: https://lore.kernel.org/r/20260715-b4-loop-partition-cleanup-v1-1-b9f59910cd1e@amutable.com
Link: https://lore.kernel.org/oe-lkp/202607150754.b660f5b9-lkp@intel.com

### DIFF
--- a/tests/loop/013
+++ b/tests/loop/013
@@ -1,0 +1,79 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 Daan De Meyer
+#
+# Verify that partitions added manually with BLKPG do not remain after a loop
+# device without LO_FLAGS_PARTSCAN is detached and reopened.
+#
+# Regression test for commit c4f4c0fc551c ("loop: remove manually added
+# partitions on detach").
+
+. tests/loop/rc
+
+DESCRIPTION="remove manually added partitions before loop device reuse"
+QUICK=1
+
+loop_device=
+udevd_stopped=
+
+requires() {
+	_have_program sfdisk
+	_have_program partx
+}
+
+cleanup() {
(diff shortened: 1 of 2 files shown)


tests/loop/013
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 Daan De Meyer
+#
+# Verify that partitions added manually with BLKPG do not remain after a loop
+# device without LO_FLAGS_PARTSCAN is detached and reopened.
+
```
dwsuse, line 10: Just a small wish from those who have to deal with backports: Would you mind to add the git commit reference which fixes the problem, e.g.

```
# Regression test for commits c4f4c0fc551c ("loop: remove manually added partitions on detach") 
```

dwsuse:
FWIW, 

Reviewed-by: Daniel Wagner <dwagner@suse.de>

kawasaki:
The new test case looks nice. I confirmed the test case passes with v7.2-rc4 kernel, and fails with v7.2-rc3 kernel. Instead of merging this PR, I applied the commit with Daniel's Reviewed-by tag. Thanks!

PR #251, opened 2026-07-16
block/045,046: add bio read-bounce alignment regression tests

0wnerDied:
## Summary

Add two regression tests for block-layer bvec alignment:

- `block/045` covers the boundary where alignment removes the only extracted bvec. It expects the direct read to fail with `EFAULT` without an out-of-bounds access.
- `block/046` covers the bounced direct-read double unpin. It keeps an io_uring fixed-buffer registration active and verifies that the registered page is not freed prematurely.

Both tests create and remove their own `null_blk` device, so no external test device is required.

## Background

Linux commit `e7b8b3c5b2a65595d506ffedafac66f0a11fbdc2` (`block: align down bounces bios`) introduced the bounced-read double unpin. Commit `df308a14585649f7683a560f5c94978ab4f8224d` (`block: fix aligning of bounced dio read bios`) fixes it.

`block/045` preserves a separate full-trim boundary case found while testing an intermediate revision of that fix. The intermediate loop accessed memory before the bvec array after consuming the final bvec; the issue was corrected before `df308a145856` was merged. Therefore both the final commit and its parent are expected to pass this test.

For `block/046`, the test creates a memory-backed `null_blk` device, enables `queue/stable_writes` before mounting XFS, and performs the partial direct read through the XFS read-bounce path.

The helper records the fixed-buffer page's PFN before the read. After triggering the partial extraction, it removes the userspace mapping while leaving the io_uring registration active and checks that PFN through `/proc/kpageflags`. On a vulnerable kernel the double unpin allows the page to enter a free-page state; with the fix applied, the io_uring pin keeps it allocated.

## Validation

- v7.2-rc5 with `df308a145856`: revised `block/046` setup passed 10/10 runs.
- v7.2-rc5 with `df308a145856` reverted: revised `block/046` setup failed 10/10 runs with `fixed-buffer page was freed (flags 0)`.
- `block/045` passed on v7.2-rc5 both with and without `df308a145856`, as expected for this boundary-only coverage.
- `block/045` passed with both normal and zoned self-configured `null_blk` devices after leaving the device at its default capacity.

### DIFF
--- a/src/bio-bounce-read.c
+++ b/src/bio-bounce-read.c
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: GPL-3.0+
+/* Copyright (C) 2026 0wnerD1ed */
+
+#define _GNU_SOURCE
+
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/io_uring.h>
+#include <linux/kernel-page-flags.h>
+#include <limits.h>
+#include <sched.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <sys/uio.h>
+#include <unistd.h>
+
+static size_t page_size;
+
+#define PAGEMAP_PRESENT (1ULL << 63)
+#define PAGEMAP_PFN_MASK ((1ULL << 55) - 1)
+
+static void *map(size_t size, int prot)
+{
+	void *p = mmap(NULL, size, prot, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+	if (p == MAP_FAILED)
+		err(EXIT_FAILURE, "mmap");
+	return p;
+}
+
+static uint64_t read_u64(int fd, off_t offset, const char *name)
+{
+	uint64_t value;
+
+	if (pread(fd, &value, sizeof(value), offset) != sizeof(value))
+		err(EXIT_FAILURE, "%s", name);
+	return value;
+}
+
+static uint64_t pagemap_pfn(int fd, const void *p)
+{
+	uint64_t entry;
+	off_t offset = (uintptr_t)p / page_size * sizeof(entry);
+
+	entry = read_u64(fd, offset, "pagemap");
+	if (!(entry & PAGEMAP_PRESENT) || !(entry & PAGEMAP_PFN_MASK))
+		errx(EXIT_FAILURE, "pagemap PFN unavailable");
+	return entry & PAGEMAP_PFN_MASK;
+}
+
+static void pin_cpu(void)
+{
+	cpu_set_t set;
+	int cpu = sched_getcpu();
+
+	if (cpu < 0)
+		err(EXIT_FAILURE, "sched_getcpu");
+	CPU_ZERO(&set);
+	CPU_SET(cpu, &set);
+	if (sched_setaffinity(0, sizeof(set), &set))
+		err(EXIT_FAILURE, "sched_setaffinity");
+}
+
+static int bounce_read(const char *size, const char *path)
+{
+	struct io_uring_params params = {};
+	void *first, *middle, *bad;
+	struct iovec fixed, iov[3];
+	uint64_t first_pfn, flags;
+	unsigned long block_size;
+	int memfd, pagemap_fd, kpageflags_fd, ring_fd, fd;
+	ssize_t ret;
+
+	block_size = strtoul(size, NULL, 10);
+	if (!block_size || block_size > UINT_MAX || block_size > page_size ||
+	    (block_size & (block_size - 1)))
+		errx(EXIT_FAILURE, "invalid block size");
+	pin_cpu();
+
+	memfd = syscall(SYS_memfd_create, "bio-bounce", 0);
+	if (memfd < 0 || ftruncate(memfd, page_size))
+		err(EXIT_FAILURE, "memfd");
+	first = mmap(NULL, page_size, PROT_READ | PROT_WRITE, MAP_SHARED, memfd,
+		     0);
+	if (first == MAP_FAILED)
+		err(EXIT_FAILURE, "mmap");
+	middle = map(block_size, PROT_READ | PROT_WRITE);
+	bad = map(block_size, PROT_NONE);
+	memset(first, 'A', page_size);
+	memset(middle, 'B', block_size);
+
+	pagemap_fd = open("/proc/self/pagemap", O_RDONLY);
+	if (pagemap_fd < 0)
+		err(EXIT_FAILURE, "pagemap");
+	first_pfn = pagemap_pfn(pagemap_fd, first);
+	kpageflags_fd = open("/proc/kpageflags", O_RDONLY);
+	if (kpageflags_fd < 0)
+		err(EXIT_FAILURE, "kpageflags");
+
+	ring_fd = syscall(SYS_io_uring_setup, 2, &params);
+	if (ring_fd < 0)
+		err(EXIT_FAILURE, "io_uring_setup");
+	fixed = (struct iovec){ first, page_size };
+	if (syscall(SYS_io_uring_register, ring_fd, IORING_REGISTER_BUFFERS,
+		    &fixed, 1) < 0)
+		err(EXIT_FAILURE, "io_uring_register");
+
+	iov[0] = (struct iovec){ first, 1 };
+	iov[1] = (struct iovec){ middle, block_size };
+	iov[2] = (struct iovec){ bad, block_size - 1 };
+	fd = open(path, O_RDONLY | O_DIRECT);
+	if (fd < 0)
+		err(EXIT_FAILURE, "open %s", path);
+	errno = 0;
+	ret = preadv(fd, iov, 3, 0);
+	if (ret != -1 || errno != EFAULT)
+		errx(EXIT_FAILURE, "preadv returned %zd (errno %d)", ret, errno);
+	close(fd);
+	munmap(first, page_size);
+	close(memfd);
+
+	flags = read_u64(kpageflags_fd, first_pfn * sizeof(flags),
+			 "kpageflags");
+	/*
+	 * The fixed-buffer pin must keep this page allocated. A page freed to a
+	 * Per-CPU Pageset (PCP) has no kpageflags, while a page drained to the
+	 * buddy allocator has KPF_BUDDY set. Either state shows that the page
+	 * was freed prematurely.
+	 */
+	if (!flags || (flags & (1ULL << KPF_BUDDY)))
+		errx(EXIT_FAILURE, "fixed-buffer page was freed (flags %#llx)",
+		     (unsigned long long)flags);
+
+	close(ring_fd);
+	close(kpageflags_fd);
+	close(pagemap_fd);
+	return EXIT_SUCCESS;
+}
+
+int main(int argc, char **argv)
+{
+	long size;
+
+	if (argc != 3)
+		return EXIT_FAILURE;
+	size = sysconf(_SC_PAGESIZE);
+	if (size <= 0)
+		errx(EXIT_FAILURE, "invalid page size");
+	page_size = size;
+	return bounce_read(argv[1], argv[2]);
+}
--- a/tests/block/045
+++ b/tests/block/045
@@ -1,0 +1,34 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 0wnerD1ed
+#
+# Exercise the case where alignment removes the only extracted bvec. This
+# caught an out-of-bounds access in an intermediate revision of the patch that
+# became df308a145856 ("block: fix aligning of bounced dio read bios").
+
+. tests/block/rc
+. common/null_blk
+
+DESCRIPTION="test direct I/O full-trim bvec alignment"
+QUICK=1
+CAN_BE_ZONED=1
+
+requires() {
+	_have_null_blk
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+
+	if ! _configure_null_blk nullb1 power=1; then
+		echo "configuring null_blk failed"
+		return 1
+	fi
+
+	if ! src/bio-full-trim /dev/nullb1; then
+		echo "bio-full-trim helper failed"
+	fi
+
+	_exit_null_blk
+	echo "Test complete"
+}
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,3 +1,4 @@
+/bio-full-trim
 /dio-offsets
 /discontiguous-io
 /loblksize
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,3 +1,4 @@
+/bio-bounce-read
 /bio-full-trim
 /dio-offsets
 /discontiguous-io
--- a/src/Makefile
+++ b/src/Makefile
@@ -13,6 +13,7 @@ HAVE_C_DEF = $(shell if echo -e "$(H)include <$(1)>\n$(H)ifdef $(2)\nHAVE_$(2)\n
 		then echo 1;else echo 0; fi)
 
 C_TARGETS := \
+	bio-full-trim \
 	dio-offsets \
 	loblksize \
 	loop_change_fd \
--- a/src/Makefile
+++ b/src/Makefile
@@ -13,6 +13,7 @@ HAVE_C_DEF = $(shell if echo -e "$(H)include <$(1)>\n$(H)ifdef $(2)\nHAVE_$(2)\n
 		then echo 1;else echo 0; fi)
 
 C_TARGETS := \
+	bio-bounce-read \
 	bio-full-trim \
 	dio-offsets \
 	loblksize \
--- a/src/bio-full-trim.c
+++ b/src/bio-full-trim.c
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: GPL-3.0+
+/* Copyright (C) 2026 0wnerD1ed */
+
+#define _GNU_SOURCE
+
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/fs.h>
+#include <stdlib.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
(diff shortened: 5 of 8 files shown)


src/bio-bounce.c
```diff
@@ -0,0 +1,171 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
kawasaki: Blktests files have SPDX-License-Identifier "GPL-3.0+". I suggest to reword the line above for consistency. Same comment for other files.

src/bio-bounce.c
```diff
@@ -0,0 +1,171 @@
...
+	off_t offset = (uintptr_t)p / page_size * sizeof(entry);
+
+	if (pread(fd, &entry, sizeof(entry), offset) != sizeof(entry))
+		err(EXIT_FAILURE, "pagemap");
+	if (!(entry & PAGEMAP_PRESENT) || !(entry & PAGEMAP_PFN_MASK))
+		errx(EXIT_FAILURE, "pagemap PFN unavailable");
+	return entry & PAGEMAP_PFN_MASK;
+}
+
+static int full_trim(const char *path)
```
kawasaki: Is there any reason to have both "full-trim" and "bounce-read" test operations in a single file? This full_trim() function looks independent from bounce_read(). Also, these two tests have different test target: block device for full-trim, and a file on XFS for bounce-read. IMO, it's the better to separate this .c file for the two tests.

tests/block/045
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0-or-later
+# Copyright (C) 2026 0wnerD1ed
+#
+# Regression test for patch "block: fix aligning of bounced dio read bios".
+# Exercise the case where alignment removes the only extracted bvec.
```
kawasaki: I assume that this test case also has a kernel side commit for fix. If so, it will be valuable to note the commit's git hash id here.

kawasaki (changes requested):
@0wnerDied Thanks for this PR. The two new test cases look valuable to me :) Please find the minor comments I made, and see if they make sense for you.

kawasaki:
Hi @0wnerDied , thanks for the patch. I have two general comments on this PR.

1. The two new test cases have single commit. I suggest to separate them into two commits, to follow "single commit - single purpose" principle.
2. Many of Linux sub-system developers prefer e-mail based reviews to GitHub pull requests. I suggest to post the commit to linux-block mailing list so that the key developers can review easily. If you prefer GitHub PR, it is still fine. In that case, I will post an e-mail to linux-block list to get their attention.

Thanks again for the contribution!

0wnerDied:
Thanks for the review. I have updated the PR accordingly:

- Changed the SPDX identifiers to GPL-3.0+.
- Split the two helpers into bio-full-trim.c and bio-bounce-read.c.
- Added the mainline fix commit, df308a145856 ("block: fix aligning of bounced dio read bios"), to both test descriptions.
- Split the tests into two separate signed commits.

I prefer to keep the GitHub PR. I would appreciate it if you could notify the linux-block mailing list for wider review.

Thanks again.

kawasaki:
Hi @0wnerDied, thank you for updating the patches.

I tried to run the new test cases.

- I ran the block/046 on v7.2-rc5 reverting the kernel commit df308a145856, and it passed. Is this expected?
- I ran the block/046 on v7.2-rc5 kernel, and it passes. Is this expected?

Now the two test cases implement test_device(). I'm guessing the test cases may be implemented with test() instead, using "_configure_null_blk nullb1": e.g., block/038. If we can do this, blktests users can run the test cases on systems that does not have real block devices. What do you think?

0wnerDied:
Hi @kawasaki, thanks for testing this and for catching the problem.

> I tried to run the new test cases.
> 
> * I ran the block/046 on v7.2-rc5 reverting the kernel commit df308a145856, and it passed. Is this expected?
> * I ran the block/046 on v7.2-rc5 kernel, and it passes. Is this expected?

Passing on v7.2-rc5 is expected. Passing after reverting df308a145856 was not expected; it showed that my previous oracle was too weak. It tried only 32 anonymous allocations and could miss the prematurely freed fixed-buffer page.

I have replaced that allocation spray with a direct per-PFN state check. The helper records the fixed-buffer PFN, triggers the partial bounced read, drops the userspace mapping while keeping the io_uring registration active, and then checks the PFN through /proc/kpageflags. On the vulnerable kernel the page has already entered a free-page state, while df308a145856 keeps it allocated through the io_uring pin.

> Now the two test cases implement test_device(). I'm guessing the test cases may be implemented with test() instead, using "_configure_null_blk nullb1": e.g., block/038. If we can do this, blktests users can run the test cases on systems that does not have real block devices. What do you think?

Both tests now implement test() and create nullb1 themselves. block/046 uses a memory-backed null_blk device and sets queue/stable_writes to 1 before mounting XFS, so it reaches the bounced direct-read path without requiring a real test device.

I validated the revised block/046 setup ten times on each kernel:

- v7.2-rc5 with df308a145856: 10/10 passed
- v7.2-rc5 with df308a145856 reverted: 10/10 failed with "fixed-buffer page was freed (flags 0)"

block/045 also passed with its self-configured null_blk device. The source build, shell syntax checks, and make check with ShellCheck 0.11.0 all pass.

I force-pushed the updated two-commit series as 8444e0b and bbeaca4. Thanks again for the useful review.

kawasaki:
Thanks for the fix. I also confirmed that block/046 now fails with v7.2-rc5 with df308a145856 reverted, and block/046 passes with v7.2-rc5 kernel. Goodl.

As to the block/045, I observe two problems:

1. When RUN_ZONED_TESTS=1 is specified in config, the test case fails to setup null_blk, because the size=16 is smaller than the default zone size 256. I suggest to drop the option size=16.
2. block/045 passes even with v7.2-rc5 with df308a145856 reverted. Is this same in your environment?

src/bio-bounce-read.c
```diff
@@ -0,0 +1,150 @@
...
+	ret = preadv(fd, iov, 3, 0);
+	if (ret != -1 || errno != EFAULT)
+		errx(EXIT_FAILURE, "preadv returned %zd (errno %d)", ret, errno);
+	close(fd);
+	munmap(first, page_size);
+	close(memfd);
+
+	flags = read_u64(kpageflags_fd, first_pfn * sizeof(flags),
+			 "kpageflags");
+	/* PCP free pages have no flags; buddy-list pages set KPF_BUDDY. */
```
kawasaki: I think this flags check is one of the keys of the test. How about to add some more description here? "Unexpected flag indicates that the fixed-buffer was freed". Also, spell out of PCP (Per-CPU Pagesets) might be kind for newbies.

0wnerDied:
> Thanks for the fix. I also confirmed that block/046 now fails with v7.2-rc5 with df308a145856 reverted, and block/046 passes with v7.2-rc5 kernel. Goodl.
> 
> As to the block/045, I observe two problems:
> 
> 1. When RUN_ZONED_TESTS=1 is specified in config, the test case fails to setup null_blk, because the size=16 is smaller than the default zone size 256. I suggest to drop the option size=16.
> 2. block/045 passes even with v7.2-rc5 with df308a145856 reverted. Is this same in your environment?

Thx for checking both cases.

Yep your zoned null_blk observation is correct. With RUN_ZONED_TESTS=1, _configure_null_blk adds zoned=1, while size=16 is smaller than null_blk's default 256 MiB zone size. I'll drop size=16 and let this non-memory-backed test use the default null_blk capacity. 

Also I reproduced your second result. block/045 passes on my v7.2-rc5 kernel with df308a145856 reverted as well. This is expected, although the current description does not explain it clearly. It's not intended to distinguish the original implementation from df308a145856. The original do-while loop already handled the single-bvec full-trim case correctly. The test covers a boundary regression that appeared in an intermediate revision of the fix: after consuming the final bvec, that loop evaluated the next condition through a pointer before the bvec array and triggered KASAN. The condition was corrected before df308a145856 was merged.

After I reported that result while testing the fix series, Christoph asked me to contribute the reproducers to blktests. So I think I need to keep block/045 as coverage for that full-trim boundary, but reword its description and commit message so they do not imply that the parent of df308a145856 should fail.

I'll do changes this day later :-)

src/bio-bounce-read.c
0wnerDied: Ah, agreed. This check should explain both the invariant and what the two flag states mean. I'll expand the comment later.

kawasaki:
Thanks for the clarification. I reverted the commit df308a145856 from v7.2-rc5 kernel, then applied [the v1 fix patch](https://lore.kernel.org/linux-block/20260713152912.2333321-1-hch@lst.de/). With this, I confirmed the block/045 fails:

```
[   90.120815] [   T1076] run blktests block/045 at 2026-08-02 15:05:55
[   90.203763] [   T1076] null_blk: disk nullb1 created
[   90.260941] [   T1096] ==================================================================
[   90.262170] [   T1096] BUG: KASAN: stack-out-of-bounds in bio_iov_iter_align_down+0x23b/0x310
[   90.263482] [   T1096] Read of size 4 at addr ffff88813efe77d8 by task bio-full-trim/1096

[   90.265039] [   T1096] CPU: 2 UID: 0 PID: 1096 Comm: bio-full-trim Not tainted 7.2.0-rc5+ #647 PREEMPT(full)
[   90.265043] [   T1096] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.17.0-10.fc44 06/10/2025
[   90.265047] [   T1096] Call Trace:
[   90.265049] [   T1096]  <TASK>
[   90.265051] [   T1096]  dump_stack_lvl+0x6a/0x90
[   90.265059] [   T1096]  print_report+0x14b/0x4b0
[   90.265063] [   T1096]  ? __virt_addr_valid+0x150/0x4d0
[   90.265067] [   T1096]  ? __virt_addr_valid+0x228/0x4d0
[   90.265071] [   T1096]  kasan_report+0xd5/0x100
[   90.265075] [   T1096]  ? bio_iov_iter_align_down+0x23b/0x310
[   90.265079] [   T1096]  ? bio_iov_iter_align_down+0x23b/0x310
[   90.265085] [   T1096]  bio_iov_iter_align_down+0x23b/0x310
[   90.265090] [   T1096]  __blkdev_direct_IO_simple+0x469/0x830
[   90.265093] [   T1096]  ? __pfx_perf_event_mmap+0x10/0x10
[   90.265098] [   T1096]  ? __pfx___blkdev_direct_IO_simple+0x10/0x10
[   90.265106] [   T1096]  ? __lock_acquire+0x3d2/0x2440
[   90.265113] [   T1096]  ? __lock_acquire+0x3d2/0x2440
[   90.265116] [   T1096]  ? __lock_acquire+0x3d2/0x2440
[   90.265120] [   T1096]  blkdev_direct_IO+0x9d0/0x1e90
[   90.265123] [   T1096]  ? reacquire_held_locks+0xe2/0x230
[   90.265126] [   T1096]  ? __mark_inode_dirty+0xaec/0x10f0
[   90.265130] [   T1096]  ? do_raw_spin_lock+0x12d/0x280
[   90.265133] [   T1096]  ? lock_acquire+0x1a9/0x320
[   90.265135] [   T1096]  ? find_held_lock+0x2b/0x80
[   90.265139] [   T1096]  ? __mark_inode_dirty+0x62d/0x10f0
[   90.265142] [   T1096]  ? __pfx_blkdev_direct_IO+0x10/0x10
[   90.265146] [   T1096]  ? lock_is_held_type+0xf6/0x1b0
[   90.265150] [   T1096]  ? rcu_lockdep_current_cpu_online+0x39/0x180
[   90.265154] [   T1096]  ? rcu_read_lock_any_held+0x66/0xa0
[   90.265157] [   T1096]  ? touch_atime+0x2c9/0x620
[   90.265162] [   T1096]  blkdev_read_iter+0x223/0x440
[   90.265166] [   T1096]  vfs_read+0x6f4/0xb30
[   90.265169] [   T1096]  ? do_raw_spin_unlock+0x55/0x230
[   90.265173] [   T1096]  ? __pfx_vfs_read+0x10/0x10
[   90.265184] [   T1096]  __x64_sys_pread64+0x18d/0x250
[   90.265187] [   T1096]  ? __pfx___x64_sys_pread64+0x10/0x10
[   90.265191] [   T1096]  ? rcu_read_unlock+0x17/0x60
[   90.265194] [   T1096]  ? lock_release+0x1b5/0x340
[   90.265197] [   T1096]  do_syscall_64+0xdf/0x790
[   90.265201] [   T1096]  ? rcu_read_unlock+0x1c/0x60
[   90.265203] [   T1096]  ? do_fault+0x711/0x1320
[   90.265206] [   T1096]  ? rcu_read_unlock+0x17/0x60
[   90.265208] [   T1096]  ? lock_release+0x1b5/0x340
[   90.265213] [   T1096]  ? __handle_mm_fault+0x10e9/0x1d50
[   90.265219] [   T1096]  ? __lock_acquire+0x3d2/0x2440
[   90.265222] [   T1096]  ? __pfx___css_rstat_updated+0x10/0x10
[   90.265228] [   T1096]  ? lock_acquire+0x199/0x320
[   90.265230] [   T1096]  ? count_memcg_events_mm.constprop.0+0x22/0x130
[   90.265233] [   T1096]  ? rcu_is_watching+0x11/0xb0
[   90.265236] [   T1096]  ? count_memcg_events+0x107/0x4c0
[   90.265239] [   T1096]  ? find_held_lock+0x2b/0x80
[   90.265242] [   T1096]  ? rcu_read_unlock+0x17/0x60
[   90.265245] [   T1096]  ? lock_release+0x1b5/0x340
[   90.265248] [   T1096]  ? find_held_lock+0x2b/0x80
[   90.265252] [   T1096]  ? exc_page_fault+0x94/0x140
[   90.265254] [   T1096]  ? lock_release+0x1b5/0x340
[   90.265258] [   T1096]  ? rcu_is_watching+0x11/0xb0
[   90.265262] [   T1096]  ? trace_hardirqs_on+0x14/0x190
[   90.265264] [   T1096]  ? preempt_count_add+0x7f/0x190
[   90.265267] [   T1096]  ? do_syscall_64+0x5d/0x790
[   90.265270] [   T1096]  ? do_syscall_64+0x8d/0x790
[   90.265272] [   T1096]  ? irqentry_exit+0xfc/0x810
[   90.265277] [   T1096]  entry_SYSCALL_64_after_hwframe+0x76/0x7e
[   90.265280] [   T1096] RIP: 0033:0x7fa696d8a08e
[   90.265283] [   T1096] Code: 4d 89 d8 e8 94 bd 00 00 4c 8b 5d f8 41 8b 93 08 03 00 00 59 5e 48 83 f8 fc 74 11 c9 c3 0f 1f 80 00 00 00 00 48 8b 45 10 0f 05 <c9> c3 83 e2 39 83 fa 08 75 e7 e8 03 ff ff ff 0f 1f 00 f3 0f 1e fa
[   90.265286] [   T1096] RSP: 002b:00007ffd9cd63770 EFLAGS: 00000202 ORIG_RAX: 0000000000000011
[   90.265291] [   T1096] RAX: ffffffffffffffda RBX: 0000000000001000 RCX: 00007fa696d8a08e
[   90.265293] [   T1096] RDX: 0000000000000200 RSI: 00007fa696f24fff RDI: 0000000000000003
[   90.265294] [   T1096] RBP: 00007ffd9cd63780 R08: 0000000000000000 R09: 0000000000000000
[   90.265296] [   T1096] R10: 0000000000000000 R11: 0000000000000202 R12: 00007ffd9cd63938
[   90.265297] [   T1096] R13: 0000000000000002 R14: 0000000000000003 R15: 00007fa696f24000
[   90.265305] [   T1096]  </TASK>

[   90.335030] [   T1096] The buggy address belongs to stack of task bio-full-trim/1096
[   90.336475] [   T1096]  and is located at offset 248 in frame:
[   90.337631] [   T1096]  __blkdev_direct_IO_simple+0x0/0x830

[   90.339416] [   T1096] This frame has 2 objects:
[   90.340374] [   T1096]  [48, 184) 'bio'
[   90.340377] [   T1096]  [256, 320) 'inline_vecs'

[   90.342786] [   T1096] The buggy address belongs to the physical page:
[   90.343958] [   T1096] page: refcount:0 mapcount:0 mapping:0000000000000000 index:0x0 pfn:0x13efe7
[   90.345446] [   T1096] flags: 0x17ffffc0000000(node=0|zone=2|lastcpupid=0x1fffff)
[   90.346752] [   T1096] raw: 0017ffffc0000000 0000000000000000 ffffea0004fbf9c8 0000000000000000
[   90.348125] [   T1096] raw: 0000000000000000 0000000000000000 00000000ffffffff 0000000000000000
[   90.349513] [   T1096] page dumped because: kasan: bad access detected

[   90.351294] [   T1096] Memory state around the buggy address:
[   90.352380] [   T1096]  ffff88813efe7680: 00 00 00 00 00 00 00 00 00 00 00 00 f1 f1 f1 f1
[   90.353729] [   T1096]  ffff88813efe7700: f1 f1 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[   90.355103] [   T1096] >ffff88813efe7780: 00 00 00 f2 f2 f2 f2 f2 f2 f2 f2 f2 00 00 00 00
[   90.356514] [   T1096]                                                     ^
[   90.357748] [   T1096]  ffff88813efe7800: 00 00 00 00 f3 f3 f3 f3 00 00 00 00 00 00 00 00
[   90.359137] [   T1096]  ffff88813efe7880: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[   90.360531] [   T1096] ==================================================================
[   90.362059] [   T1096] Disabling lock debugging due to kernel taint
[   90.733507] [   T1123] null_blk: disk nullb0 created
[   90.734825] [   T1123] null_blk: module loaded
[   90.852660] [   T1076] run blktests block/045 at 2026-08-02 15:05:55
[   90.887280] [   T1076] null_blk: nullb1: using native zone append
[   90.896279] [   T1076] null_blk: disk nullb1 created
[  328.965251] [   T1236] null_blk: disk nullb0 created
[  328.966506] [   T1236] null_blk: module loaded
```

IOW, I confirmed that block/045 confirms the real bug which appeared during the development of the kernel commit df308a145856.

kawasaki:
@0wnerDied I merged this PR. Thank you again for the contribution!

PR #252, opened 2026-07-22
docs: add fully annotated configuration example

dwsuse:
Adds config.example with detailed comments for all required and optional settings to make initial setup easier for new users.

### DIFF
--- a/config.example
+++ b/config.example
@@ -1,0 +1,169 @@
+# blktests example configuration file
+#
+# Copy this file to "config" (or pass it with "./check -c config.example")
+# and adjust the settings for your environment.
+#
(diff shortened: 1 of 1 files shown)


kawasaki:
Thanks! It was my long due, and now I'm happy to have the config.example.

Just a side not: during review I found that the three config parameters OUTPUT, LOGGER_PROG and MODPROBE_PATIENT_RM_TIMEOUT_SECONDS are in the config.example, but not documented in Documentation/running-tests.md. It might be add descriptions for them in running-tests.md for completeness.

igaw:
I didn't know them either, but copilot found them :)

PR #253, opened 2026-07-27
Add ability to have multiple host ports

Mr-Bossman:
# Add various features namely the ability to have multiple host ports 

This will be used in my upcoming PR for testing nvme marginal support.
All the commits are independent of one-anther, if anyone thinks its a good idea
to split them into multiple PRs I can .

### makefile: Add rules to generate ctags
    
    If cscope and ctags are installed create the databases needed
    to browse these shell scripts with cscope/ctags.

### nvme: Add ability to have multiple host ports
    
    Currently, the nvme tests only support a single host port.
    This patch adds the ability to have multiple host ports, which will
    more accuratly reflect a real-world multipath scenario.

### nvme: Add _nvmet_target_add_ports
    
    Add `_nvmet_target_add_ports` to allow for more ports to be added after
    `_nvmet_target_setup` has been called.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,12 +6,23 @@ all:
 
 clean:
 	$(MAKE) -C src clean
+	rm -f tags cscope.*
 
 install:
 	install -m755 -d $(dest)
 	install check $(dest)
 	cp -R tests common $(dest)
 	$(MAKE) -C src dest=$(dest)/src install
+
+tags: check_utility
+	@rm -f cscope.* tags
+	find . -type f -print | grep -v -E ".out|.git|.md|.files|.cmd|tags|LICENSES|.el|.jinja2|.py|results" > cscope.files
+	ctags -R --language-force=sh -L cscope.files --extras=+f
+	cscope -bq -I/usr/include
+
+check_utility:
+	@command -v ctags &>/dev/null || (echo "Error: 'ctags' not found in PATH." && exit 1)
+	@command -v cscope &>/dev/null || (echo "Error: 'cscope' not found in PATH." && exit 1)
 
 # SC2119: "Use foo "$@" if function's $1 should mean script's $1". False
 # positives on helpers like _init_scsi_debug.
@@ -36,4 +47,4 @@ check-parallel:
 	find -L -name '*.out' -perm /u=x+g=x+o=x -printf '%p is executable\n' | grep . && ret=1; \
 	exit $$ret
 
-.PHONY: all check check-parallel install
+.PHONY: all check check-parallel install tags check_utility
--- a/common/nvme
+++ b/common/nvme
@@ -1102,6 +1102,85 @@ _nvmet_target_setup() {
 			"${hostkey}" "${ctrlkey}"
 }
 
+_nvmet_target_add_ports() {
+	local ctrlkey=""
+	local hostkey=""
+	local subsysnqn="${def_subsysnqn}"
+	local subsys_uuid
+	local port p
+	local num_ports=1
+	local tls="none"
+	local cfs_path
+
+	while [[ $# -gt 0 ]]; do
+		case $1 in
+			--ctrlkey)
+				ctrlkey="$2"
+				shift 2
+				;;
+			--hostkey)
+				hostkey="$2"
+				shift 2
+				;;
+			--subsysnqn)
+				subsysnqn="$2"
+				shift 2
+				;;
+			--subsys-uuid)
+				subsys_uuid="$2"
+				shift 2
+				;;
+			--ports)
+				num_ports="$2"
+				shift 2
+				;;
+			--tls)
+				tls="not-required"
+				shift 1
+				;;
+			--force-tls)
+				tls="required"
+				shift 1
+				;;
+			*)
+				echo "WARNING: unknown argument: $1"
+				shift
+				;;
+		esac
+	done
+
+	if [[ -n "${nvme_target_control}" ]]; then
+		if [[ -n "${hostkey}" ]]; then
+			ARGS+=(--hostkey "${hostkey}")
+		fi
+		if [[ -n "${ctrlkey}" ]]; then
+			ARGS+=(--ctrkey "${ctrlkey}")
+		fi
+
+		"${nvme_target_control}" add-ports \
+					 --ports "${num_ports}" \
+					 --subsysnqn "${subsysnqn}" \
+					 --subsys-uuid "${subsys_uuid:-$def_subsys_uuid}" \
+					 --hostnqn "${def_hostnqn}" \
+					 "${ARGS[@]}" &> /dev/null
+		echo "'${nvme_target_control} add-ports' is a stub"
+		return
+	fi
+
+	cfs_path="${NVMET_CFS}/subsystems/${subsystem}"
+	if [[ ! -d "${cfs_path}" ]]; then
+		echo "Can not add ports to a non-existent subsystem. subsystem not found"
+		return 1;
+	fi
+
+	p=0
+	while (( p < num_ports )); do
+		port="$(_create_nvmet_port ${tls})"
+		_add_nvmet_subsys_to_port "${port}" "${subsysnqn}"
+		p=$(( p + 1 ))
+	done
+}
+
 _nvmet_target_cleanup() {
 	local ports
 	local port
--- a/common/nvme
+++ b/common/nvme
@@ -30,6 +30,7 @@ NVMET_CFS="/sys/kernel/config/nvmet"
 NVMET_DFS="/sys/kernel/debug/nvmet"
 nvme_trtype=${nvme_trtype:-}
 nvme_adrfam=${nvme_adrfam:-}
+declare -ag ports_to_hosts
 
 # TMPDIR can not be referred out of test() or test_device() context. Instead of
 # global variable def_flie_path, use this getter function.
@@ -150,6 +151,7 @@ _cleanup_nvmet() {
 	local subsys
 	local transport
 	local name
+	local host_port
 
 	if [[ ! -d "${NVMET_CFS}" ]]; then
 		return 0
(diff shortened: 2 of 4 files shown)


common/nvme
```diff
@@ -1102,6 +1102,45 @@ _nvmet_target_setup() {
...
+				echo "WARNING: unknown argument: $1"
+				shift
+				;;
+		esac
+	done
+
+	p=0
+	while (( p < num_ports )); do
+		port="$(_create_nvmet_port ${tls})"
+		_add_nvmet_subsys_to_port "${port}" "${subsysnqn}"
```
igaw, line 1242: The add port command should also be forward to NVME_TARGET_CONTROL.

Mr-Bossman: Im not sure what you mean? Like call `"${nvme_target_control}" setup`?

igaw:
```
- NVME_TARGET_CONTROL: When defined, the generic target setup/cleanup code will
  be skipped and this script gets called. This makes it possible to run
  the fabric nvme tests against a real target.
```

Yes, this script is able to steer any external target: https://github.com/linux-blktests/blktests/blob/master/contrib/nvme_target_control.py

It looks like a new sub command would be necessary in the script.

(and after looking at it, I need to update the script to use ssh to a remote machine instead)

Mr-Bossman: oh ok ill look into that

common/nvme
```diff
@@ -222,6 +229,7 @@ _cleanup_nvmet() {
 }
 
 _setup_nvmet() {
+	local -i host_ports=${1:-1}
```
kawasaki: Nit: does this local variable "host_ports" mean number of host ports? If so, I suggest to rename it to "nr_host_ports". This is just my preference, though.

common/nvme
```diff
@@ -920,9 +948,20 @@ _get_nvmet_ports() {
...
 }
 
+_get_fc_host_port() {
+	local -i port=${1}
+
+	echo "${ports_2_hosts[${port}]}"
+}
+
+_get_fc_host_ports() {
+	printf "%s\n" "${ports_2_hosts[@]}" | sort -u
```
kawasaki: Blktests encourages long option over short options for readability and stability. I suggest 'sort --unique' in place of 'sort -u'.

common/nvme
```diff
@@ -1182,7 +1233,10 @@ _nvmet_target_cleanup() {
 	_remove_nvmet_host "${def_hostnqn}"
 
 	if [[ "${nvme_trtype}" == "fc" ]]; then
-		_nvme_fcloop_del_lport "${def_local_wwnn}" "${def_local_wwpn}"
+		for host_port in $(_get_fc_host_ports); do
```
kawasaki: Nit: the local variable host_port is not declared.

common/nvme
```diff
@@ -968,7 +1007,7 @@ _find_nvme_dev() {
 	for dev in /sys/class/nvme/nvme*; do
 		[ -e "$dev" ] || continue
 		dev="$(basename "$dev")"
-		subsysnqn="$(cat "/sys/class/nvme/${dev}/subsysnqn" 2>/dev/null)"
+		subsysnqn="$(cat "/sys/class/nvme/${dev}/subsysnqn" 2> /dev/null)"
```
kawasaki: This change does not look relevant to this commit. Let's follow 'single commit, single purpose" principle.

Makefile
```diff
@@ -6,13 +6,24 @@ all:
...
 	$(MAKE) -C src clean
+	rm -f tags cscope.*
 
 install:
 	install -m755 -d $(dest)
 	install check $(dest)
 	cp -R tests common $(dest)
 	$(MAKE) -C src dest=$(dest)/src install
 
+tags: check_utility
```
kawasaki: A stray space at the end of this line.

common/nvme
```diff
@@ -1206,6 +1142,13 @@ _nvmet_target_cleanup() {
 	_remove_nvmet_subsystem "${subsysnqn}"
 	_remove_nvmet_host "${def_hostnqn}"
 
+	if [[ "${nvme_trtype}" == "fc" ]]; then
+		_nvme_fcloop_del_lport "${def_local_wwnn}" "${def_local_wwpn}"
+	fi
+	if [[ "${nvme_trtype}" == "rdma" ]]; then
```
kawasaki: This if block is not explained in the commit message. Why this change is required?

common/nvme
```diff
@@ -212,7 +212,7 @@ _cleanup_nvmet() {
 	trap SIGINT
 
 	if [[ "${nvme_trtype}" == "fc" ]]; then
-		_nvme_fcloop_del_lport "${def_local_wwnn}" "${def_local_wwpn}"
+		_nvme_fcloop_del_lport "${def_local_wwnn}" "${def_local_wwpn}" 2> /dev/null
```
kawasaki: This change is not describe in the commit message. Why we need to ignore _nvme_fcloop_del_lport stderr?

kawasaki (changes requested):
Hi @Mr-Bossman thank you for the PR. Overall, four out of the three commits look good. I'm not so sure about the first commit, "nvme: Delete lports in _nvmet_target_cleanup". Its commit message says that _nvmet_target_cleanup() is called when Ctrl-C is typed, but my understanding is that _cleanup_nvmet() is called when Ctrl-C is typed.

Other three commits looks valid. Please find some minor comments I made for them, and see if they make sense for you.

I have not yet done trail test runs. Will do later. Thanks!

common/nvme
Mr-Bossman: good point.

common/nvme
Mr-Bossman: good to know!

common/nvme
Mr-Bossman: You are right i think when formatting my commits i accidentally changed this line instead of the line i added in `_cleanup_nvmet`.

Makefile
Mr-Bossman: odd I thought my editor removes trailing whitespace, i guess not in this case.

common/nvme
Mr-Bossman: ah interesting i didn't know for loop variables were not local

common/nvme
Mr-Bossman: i relize the commit description is a bit misleading, but this function gets called at the end of the test to clean up the test, but it doesnt delete the `lport` or stop `soft rdma`.  Delete the `lport` and stoping `soft rdma`  is handled in `_cleanup_nvmet` which is called by the caller. ideally `_cleanup_nvmet` is only precautionary for when the test fails and doesn't clean itself up. Adding cleanup for `lport` and `rdma` here will fully cleanup the test on exit.

common/nvme
Mr-Bossman: I just added a fix in later commits, but basically there is no easy way to check if `lport` has been cleaned up or not so we just unconditionally delete it even if it doesnt exist. Sence i added `_nvme_fcloop_del_lport` to `_nvmet_target_cleanup`  the `lport` has already been removed causing this to error with `ENOENT`

Mr-Bossman:
@kawasaki I dropped the "nvme: Delete lports in _nvmet_target_cleanup" commit

common/nvme
```diff
@@ -1166,6 +1102,85 @@ _nvmet_target_setup() {
...
+			ARGS+=(--ctrkey "${ctrlkey}")
+		fi
+
+		"${nvme_target_control}" add-ports \
+					 --ports "${num_ports}" \
+					 --subsysnqn "${subsysnqn}" \
+					 --subsys-uuid "${subsys_uuid:-$def_subsys_uuid}" \
+					 --hostnqn "${def_hostnqn}" \
+					 "${ARGS[@]}" &> /dev/null
+		echo "`${nvme_target_control} add-ports` is a stub"
```
kawasaki: This line caused "make shellcheck" failure: [ref](be47c8b6)
I suggest to replace the back ticks ` with single  quote'.

contrib/nvme_target_control.py
```diff
@@ -117,6 +117,11 @@ def target_setup(args):
     subprocess.call(['python3', nvmetcli, '--remote=' + remote,
                      'restore', args.subsysnqn + '.json'])
 
+def target_add_ports(_):
```
kawasaki, line 120: I suggest to mention this change of contrib/nvme_target_control.py in the commit message. The added "add-ports" command is a stub at this moment.  It will be also valuable to note your outlook for the stub implementation in the commit log (will your implement it later? Or do you expect someone to implement it?)

kawasaki (changes requested):
@Mr-Bossman Thanks for the update. I did full blktests run with the changes, and did not see regression. Good from test run point of view.

The script change looks good from logic point of view. I made two comments about shellcheck WARN and the commit message contents. Please check them out.

contrib/nvme_target_control.py
Mr-Bossman: I think the commit message is fixed now.

common/nvme
```diff
@@ -30,6 +30,7 @@ NVMET_CFS="/sys/kernel/config/nvmet"
 NVMET_DFS="/sys/kernel/debug/nvmet"
 nvme_trtype=${nvme_trtype:-}
 nvme_adrfam=${nvme_adrfam:-}
+declare -ag ports_2_hosts
```
igaw: Just wondering what `ports_2_hosts` means. Should it express the mapping of the ports to hosts, or is it something else. If it is the mapping, please use `ports_to_hosts` or something similar.

igaw:
Looks good, just the name of `hosts_2_ports` is not clear to me. 

```
Reviewed-by: Daniel Wagner <dwagner@suse.de>
```

common/nvme
Mr-Bossman: it is a mapping ill change it to `ports_to_hosts`

kawasaki:
@Mr-Bossman The updated commits look good to me. I have applied the three commits manually to leave Daniel's reviewd-by tags in each commit message. Let me close this PR. Thanks!

Mr-Bossman:
@kawasaki Thanks so much!

PR #254, opened 2026-07-27
nvme/{014,019}: Print the return value instead of shell execution

dirkmueller:


kawasaki (changes requested):
Hi @dirkmueller Thank you for catching this bug! The change looks good to me.

May I ask you to update the commit message?

1. Please add title with valid prefix. I suggest "nvme/{014,019}: Print the return value instead of shell execution
2. Please add your Signed-off-by tag as guided in [CONTRIBUTING.md](https://github.com/linux-blktests/blktests/blob/master/CONTRIBUTING.md).

Thanks!

kawasaki:
I have merged this PR, thanks!

PR #256, opened 2026-07-29
Md tests

schlad:


tests/md/006
```diff
@@ -0,0 +1,53 @@
...
+	_have_brd
+	_have_fio
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+
+	local mddev="/dev/md/blktests_md_${TEST_NAME##*/}"
+
+	if ! _init_brd rd_size=524288 rd_nr=2; then
```
igaw, line 26: from past experience, usually it's better to calculate the values not hard code them in case they are depended on the drive.

schlad: good point.
I looked how md/001 is doing that and I did that in the similar way, but yeah I agree.

igaw: yeah, sorry, sometimes you just start pulling a thread....

kawasaki:
Thanks, I took a look in the commits and they look good overall. Let me leave a few comments.

- As discussed, duplication in test cases from md/005 to md/009 should be avoided. I suggest to combine them into a single test case, but this is not a strong opinion. There could be other way to avoid the duplication.

- There are some short options such as "mkdir -p" or "grep -q", and I would suggest to use long options for them, but this is a nit comment.

- When I ran md/011 on v7.2-rc5 kernel, it passed at the 1st run. It failed at 2nd run with the message below:
```
md/011 => nvme0n1 nvme2n1 nvme3n1 nvme4n1 (reproduce raid1 write-vs-recovery race with concurrent I/O on ext4) [failed]
    runtime  11.084s  ...  15.817sests/blktests/results/tmpdir.md.011.3BS/mnt': Device or resource busy
    --- tests/md/011.out        2026-07-31 18:14:00.508000000 +0900
    +++ /home/shin/Blktests/blktests/results/nvme0n1_nvme2n1_nvme3n1_nvme4n1/md/011.out.bad     2026-07-31 18:29:22.746000000 +0900
    @@ -1,2 +1,6 @@
     Running md/011
    +mdadm: cannot open /dev/nvme0n1: Device or resource busy
    +timed out waiting for initial resync to finish
    +/dev/md/blktests_md_011 is mounted; will not make a filesystem here!
    +timed out waiting for recovery to start
     Test complete
```

Is this the expected failure?

I also wonder if md/011 has corresponding kernel side fix commit or not.

schlad:
> Thanks, I took a look in the commits and they look good overall. Let me leave a few comments.
> 
>     * As discussed, duplication in test cases from md/005 to md/009 should be avoided. I suggest to combine them into a single test case, but this is not a strong opinion. There could be other way to avoid the duplication.
> 
>     * There are some short options such as "mkdir -p" or "grep -q", and I would suggest to use long options for them, but this is a nit comment.
> 
>     * When I ran md/011 on v7.2-rc5 kernel, it passed at the 1st run. It failed at 2nd run with the message below:
> 
> 
> ```
> md/011 => nvme0n1 nvme2n1 nvme3n1 nvme4n1 (reproduce raid1 write-vs-recovery race with concurrent I/O on ext4) [failed]
>     runtime  11.084s  ...  15.817sests/blktests/results/tmpdir.md.011.3BS/mnt': Device or resource busy
>     --- tests/md/011.out        2026-07-31 18:14:00.508000000 +0900
>     +++ /home/shin/Blktests/blktests/results/nvme0n1_nvme2n1_nvme3n1_nvme4n1/md/011.out.bad     2026-07-31 18:29:22.746000000 +0900
>     @@ -1,2 +1,6 @@
>      Running md/011
>     +mdadm: cannot open /dev/nvme0n1: Device or resource busy
>     +timed out waiting for initial resync to finish
>     +/dev/md/blktests_md_011 is mounted; will not make a filesystem here!
>     +timed out waiting for recovery to start
>      Test complete
> ```
> 
> Is this the expected failure?
> 
> I also wonder if md/011 has corresponding kernel side fix commit or not.

Yes, the problem is known and there is a known fix. I will update the test information with the details.

Hmm I did not run it on v7.2-rc5 kernel; I will do and see why this is timing out in the 2nd run.

PR #257, opened 2026-07-29
nvme: rename DH to KX according TP4201

dwsuse:
The DH (Diffie Hellman) prefix is replaced with KX (key exchange) as defined in TP4201 Enhancements for Post-Quantum Crypto Algorithms.

Rename all blktests internal function, arguments accordingly.

nvme-cli 3 is deprecating and replaces them with sub commands
- `nvme check-dhchap-key` -> `nvme keys check-kxchap`
- `nvme check-tls-key`-> `nvme keys check-tls`
- `nvme gen-dhchap-key` -> `nvme keys gen-kxchap`
- `nvme gen-tls-key` -> `nvme keys gen-tls`
- `nvme tls-key` -> `nvme keys import`, `nvme keys export`, `nvme keys revoke`

nvme-cli 3.x also supports the new naming scheme, thus auto detect it and use the 'keys' plugin which uses the new naming scheme.

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -37,6 +37,60 @@ _nvme_def_file_path() {
 	echo "${TMPDIR}/img"
 }
 
+_nvme_kxchap_cmd_style() {
+	if [[ -z "${_nvme_kxchap_style:-}" ]]; then
+		if nvme keys gen-kxchap --nqn nvmf-test-subsys > /dev/null 2>&1; then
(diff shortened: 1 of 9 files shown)


kawasaki:
The change is valuable, and script change looks good. I also ran nvme test group with nvme-cli tags v3.0-b.3 and v3.0-b.4, and result was same for both tags. Thanks!

PR #259, opened 2026-08-06
loop: test partition cleanup when the backing file is changed

daandemeyer:
Partitions can be added to a loop device with BLKPG while
LO_FLAGS_PARTSCAN is clear. loop_change_fd() only rescanned when
LO_FLAGS_PARTSCAN was set, so once disk_force_media_change() stopped
setting GD_NEED_PART_SCAN such partitions survived the backing file swap
and kept describing the old file.

Add a regression test that adds a partition with partx to a read-only
loop device, swaps in a backing file of the same size without a partition
table, and verifies that the partition is gone afterwards. Stop an active
systemd-udevd during the test so userspace cannot hide missing kernel
cleanup, and restore it through the test cleanup hook.

Signed-off-by: Daan De Meyer <daan@amutable.com>

### DIFF
--- a/tests/loop/014
+++ b/tests/loop/014
@@ -1,0 +1,76 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 Daan De Meyer
+#
+# Verify that partitions added manually with BLKPG do not survive a
+# LOOP_CHANGE_FD on a loop device without LO_FLAGS_PARTSCAN.
+#
+# Regression test for commit "loop: drop stale partitions on LOOP_CHANGE_FD".
+
+. tests/loop/rc
+
+DESCRIPTION="remove manually added partitions when the backing file is changed"
+QUICK=1
+
+loop_device=
+udevd_stopped=
+
+requires() {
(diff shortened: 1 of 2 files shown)


kawasaki:
@daandemeyer Thanks for this PR. The code change look good to me. I will wait for the kernel patch gets upstreamed before I merge this PR.

One quick question: I guess the reason to use partx is that sfdisk does not have the feature "partx --add" provides. Is this guess correct?

daandemeyer:
> One quick question: I guess the reason to use partx is that sfdisk does not have the feature "partx --add" provides. Is this guess correct?

I am unsure, I was notified of a regression in some project using partx. I did not take a closer look at the project itself.

kawasaki:
> > One quick question: I guess the reason to use partx is that sfdisk does not have the feature "partx --add" provides. Is this guess correct?
> 
> I am unsure, I was notified of a regression in some project using partx. I did not take a closer look at the project itself.

Thanks, I did some more investigation, but could not find the way to replace "partx --add" with sfdisk or other util-linux tools. Then the current commit looks the best.

PR #260, opened 2026-08-06
nbd: test partition cleanup after a disconnect

daandemeyer:
The disconnect path relied on disk_force_media_change() setting
GD_NEED_PART_SCAN to drop the partitions of the old device on the next
open. Once that stopped happening, nbd_clear_sock_ioctl() left them
behind. With the default max_part of 16 this stays invisible because
nbd_open() sets the bit itself on a disconnected device, so the stale
partitions are only observable with max_part=0.

Add a regression test that partitions a connected nbd device, disconnects
it over the ioctl path, reopens it and verifies that the partition is
gone. Require max_part to be 0 so the test is skipped rather than passing
vacuously in the default configuration, and stop an active systemd-udevd
during the test so userspace cannot hide missing kernel cleanup.

Signed-off-by: Daan De Meyer <daan@amutable.com>

### DIFF
--- a/tests/nbd/005
+++ b/tests/nbd/005
@@ -1,0 +1,88 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-3.0+
+# Copyright (C) 2026 Daan De Meyer
+#
+# Verify that the partitions of an nbd device do not survive NBD_CLEAR_SOCK.
+#
+# With a non-zero max_part, nbd_open() sets GD_NEED_PART_SCAN itself on a
+# disconnected device, which hides a missing teardown in the disconnect path.
+# Load nbd with max_part=0 to run this test.
+#
+# Regression test for commit "nbd: drop stale partitions on NBD_CLEAR_SOCK".
+
+. tests/nbd/rc
+
+DESCRIPTION="remove partitions of a disconnected nbd device"
+QUICK=1
+
+server_started=
+nbd_connected=
+udevd_stopped=
+
+requires() {
+	_have_nbd
+	_have_module_param_value nbd max_part 0
+	_have_program parted
+}
+
+cleanup() {
+	if [[ -n "$nbd_connected" ]]; then
+		nbd-client -nonetlink -d /dev/nbd0 &>/dev/null
+	fi
+	if [[ -n "$server_started" ]]; then
+		_stop_nbd_server
+	fi
+	if [[ -n "$udevd_stopped" ]]; then
+		_systemd_start_udevd
+	fi
+}
+
+test() {
+	echo "Running ${TEST_NAME}"
+	_register_test_cleanup cleanup
+
+	_start_nbd_server
+	server_started=1
+
+	if ! nbd-client -nonetlink -N export localhost /dev/nbd0 >>"$FULL" 2>&1 ||
+			! _wait_for_nbd_connect; then
+		echo "Failed to connect nbd device"
+		echo "Test complete"
+		return
+	fi
+	nbd_connected=1
+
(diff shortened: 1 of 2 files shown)


tests/nbd/005
```diff
@@ -0,0 +1,88 @@
...
+DESCRIPTION="remove partitions of a disconnected nbd device"
+QUICK=1
+
+server_started=
+nbd_connected=
+udevd_stopped=
+
+requires() {
+	_have_nbd
+	_have_module_param_value nbd max_part 0
```
kawasaki, line 24: IIUC, it is expected that the blktests users reload nbd module with max_part=0 option value before running this test case. Is this guess correct?

If my guess is correct, I suggest to include the user operation in this test case with the following change.

- Call "_have_module nbd" in _requires() instead of "_have_module_param_value" to ensure that nbd module is not built-in, but loadable.
- Call "modprobe --remove nbd" and "modprobe nbd max_part=0" at the beginning of the test case.
- Call "modprobe --remove nbd" in cleanup().

I think the following change works.

```
diff --git a/tests/nbd/005 b/tests/nbd/005
index 3b2899c..130a2ec 100755
--- a/tests/nbd/005
+++ b/tests/nbd/005
@@ -21,7 +21,7 @@ udevd_stopped=

 requires() {
        _have_nbd
-       _have_module_param_value nbd max_part 0
+       _have_module nbd
        _have_program parted
 }

@@ -35,12 +35,22 @@ cleanup() {
        if [[ -n "$udevd_stopped" ]]; then
                _systemd_start_udevd
        fi
+
+       modprobe --remove nbd
 }

 test() {
        echo "Running ${TEST_NAME}"
        _register_test_cleanup cleanup

+       if ! modprobe --remove --quiet nbd; then
+               return
+       fi
+
+       if ! modprobe nbd max_part=0; then
+               return
+       fi
+
        _start_nbd_server
        server_started=1
```

tests/nbd/005
```diff
@@ -0,0 +1,88 @@
...
+	if ! nbd-client -nonetlink -N export localhost /dev/nbd0 >>"$FULL" 2>&1 ||
+			! _wait_for_nbd_connect; then
+		echo "Failed to connect nbd device"
+		echo "Test complete"
+		return
+	fi
+	nbd_connected=1
+
+	{
+	parted -s /dev/nbd0 mklabel gpt
```
kawasaki, line 56: I suggest the long option "--script" instead of short option "-s" here.
Just my preference, but I suggest one more deeper indent here.
Same for the next line.

kawasaki (changes requested):

kawasaki:
Thanks for this PR.

While I ran the new test case, I observed a weird message below:

```
nbd/005 (remove partitions of a disconnected nbd device)     [passed]
    runtime  7.668s  ...  7.016s
cat: /nbd.pid: No such file or directory
tests/nbd/rc: line 87: kill: `': not a pid or valid job spec
```

Do you observe the same message?

I think this is because the clean up handler was called twice. It should be called only once, and I'm guessing this is a bug in blktestes framework. The change below avoids the weird message:

```
diff --git a/check b/check
index e7eca84..94f44dd 100755
--- a/check
+++ b/check
@@ -414,6 +414,7 @@ _register_test_cleanup() {
 _cleanup() {
        if [[ -v TEST_CLEANUP ]]; then
                ${TEST_CLEANUP}
+               unset TEST_CLEANUP
        fi

        if [[ "${TMPDIR:-}" ]]; then
```

I'm thinking to create a formal patch for this.

PR #261, opened 2026-08-06
Three small blktests improvements

bvanassche:


kawasaki:
@bvanassche Thanks, I applied the patches manually make an amendment. I replaced the grep short options with long options. Anyway, the patches are now in the master branch. Let me close this PR.

PR #262, opened 2026-08-10
nvme: Delete lports in _nvmet_target_cleanup

Mr-Bossman:
Delete the lports in `_nvmet_target_cleanup`, which is called when when the test ends. Additionally, ignore file not found errors when deleting the lports, in `_cleanup_nvmet` as they might have already been deleted by `_nvmet_target_cleanup`.

This will also fix the issue of `common/nvme: line 186: echo: write error: No such file or directory ` being printed after the test has finished which will happen when running tests with `NVMET_TRTYPES=fc ./check tests/nvme/*`

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -272,7 +272,7 @@ _cleanup_nvmet() {
 	if [[ "${nvme_trtype}" == "fc" ]]; then
 		for host_port in $(_get_fc_host_ports); do
 			_nvme_fcloop_del_lport "$(_host_wwnn "$host_port")" \
-					       "$(_host_wwpn "$host_port")"
+					       "$(_host_wwpn "$host_port")"  2> /dev/null
 		done
 	fi
 	if [[ "${nvme_trtype}" == "rdma" ]]; then
@@ -1402,6 +1402,13 @@ _nvmet_target_cleanup() {
 	done
 	_remove_nvmet_subsystem "${subsysnqn}"
 	_remove_nvmet_host "${def_hostnqn}"
+
+	if [[ "${nvme_trtype}" == "fc" ]]; then
+		_nvme_fcloop_del_lport "${def_local_wwnn}" "${def_local_wwpn}"
+	fi
+	if [[ "${nvme_trtype}" == "rdma" ]]; then
+		stop_soft_rdma
+	fi
 
 	if [[ "${blkdev_type}" == "device" ]]; then
 		_cleanup_blkdev


kawasaki:
It is still not clear for me why the change is required. My understanding is that,

- most of nvme group test cases call _setup_nvmet()
- _setup_nvmet() calls "_register_test_cleanup _cleanup_nvmet"
- then _cleanup_nvmet() is called for most of nvme group test cases regardless whether the test case passed or failed
- _cleanup_nvmet() deletes lport ( and calls stop_soft_rdma )

Then _nvmet_target_cleanup() should not need to delete lport.

I guess you needed this change to avoid some failures. What kind of failure do you observe?

Mr-Bossman:
> It is still not clear for me why the change is required. My understanding is that,
> 
> * most of nvme group test cases call _setup_nvmet()
> * _setup_nvmet() calls "_register_test_cleanup _cleanup_nvmet"
> * then _cleanup_nvmet() is called for most of nvme group test cases regardless whether the test case passed or failed
> * _cleanup_nvmet() deletes lport ( and calls stop_soft_rdma )
> 
> Then _nvmet_target_cleanup() should not need to delete lport.
> 
> I guess you needed this change to avoid some failures. What kind of failure do you observe?

For some reason when `_cleanup_nvmet` calls `_nvme_fcloop_del_lport` this error happens `common/nvme: line 186: echo: write error: No such file or directory` moving `_nvme_fcloop_del_lport` to `_nvmet_target_cleanup` fixes this issue. 

IMO `_nvmet_target_cleanup` should cleanup as much as possible and not leave stuff etxra stuff for `_cleanup_nvmet` to do which is why i also added `stop_soft_rdma`.

kawasaki:
Thanks, I'm interested in the failure. May I know how to recreate the failure? I would like to take a closer look in it.

At the blktests master branch tip, the line 186 of common/nvme is not in _nvme_fcloop_del_lport(), but in _nvme_fcloop_del_tport(). When you see the failure, line 186 is in _nvme_fcloop_del_lport(), right?

k
IMO, we should have less duplication between _cleanup_nvmet() and _nvmet_target_cleanup(). Then I would like to fully understand why the duplication is required.

Mr-Bossman:
This happens on Centos 10, same with https://github.com/linux-blktests/blktests/pull/263. sorry for the lack of updates, I have been a bit busy.

PR #263, opened 2026-08-10
README.md: Require nvme-cli >= 3.0

Mr-Bossman:
Commit df7cf50409ce ("nvme: rely on nvme-cli status code") changed nvme-cli to use the exit code to determine the result of the command, which is only available in nvme-cli >= 3.0. This commit updates the readme to match this requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The dependencies are minimal, but make sure you have them installed:
 Some tests require the following:
 
 - e2fsprogs, xfsprogs, f2fs-tools and btrfs-progs
-- nvme-cli
+- nvme-cli (>= 3.0)
 - multipath-tools (Debian, openSUSE, Arch Linux) or device-mapper-multipath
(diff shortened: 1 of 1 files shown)


igaw:
nvme-cli 2.16 also returns error codes, there is no hard dependency on 3.0.

kawasaki:
Thanks for this PR, but I have the same view as @igaw . Quite a few blktests users will keep on running blktests with nvme-cli older than 3.0, and I do not want to upset them. For that purpose, blktests should work with the older nvme-cli. If you see any failure with older nvme-cli versions, report will be appreciated.

Mr-Bossman:
> If you see any failure with older nvme-cli versions, report will be appreciated.

Yes let me see if i can make a simple example. The gist is that `nvme disconnect` for v2.16 sometimes returns 1 even if the disconnect was successful, it only happens sometimes though.

PR #266, opened 2026-08-19
nvme: Fix fio process termination on completion

Mr-Bossman:
fio was spawned inside a function, killing the function PID does not kill the fio process. Use set -m to enable job control and kill the fio process proccess group instead of the function PID.

### DIFF
--- a/tests/nvme/032
+++ b/tests/nvme/032
@@ -30,6 +30,7 @@ device_requires() {
 test_device() {
 	echo "Running ${TEST_NAME}"
 
+	local fio_pid
 	local sysfs
 	local attr
 	local m
@@ -38,8 +39,11 @@ test_device() {
 	sysfs="/sys/bus/pci/devices/${pdev}"
 
 	# start fio job
+	set -m
 	_run_fio_rand_io --filename="$TEST_DEV" \
 		--group_reporting --time_based --runtime=1d &> /dev/null &
+	set +m
+	fio_pid=$!
 
 	sleep 5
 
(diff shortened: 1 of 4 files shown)


kawasaki:
@Mr-Bossman  Thank you for this PR. It's a nice catch. The change looks good to me. I will do an evaluation run to confirm no side effect.

Two questions:

1. Do you see any failures that this commit fix? If so, it's worth describing in the commit message. Regardless of the answer to this question, I think this PR change is nice.
2. I found nvme/032 also kills the background _run_fio_rand_io(). Do you think it needs fix also? If so, would you mind creating another PR (or include in this PR)?

kawasaki:
I ran whole nvme test group and there was no regression. Good. Let me wait for the answers to my questions.

Mr-Bossman:
> @Mr-Bossman Thank you for this PR. It's a nice catch. The change looks good to me. I will do an evaluation run to confirm no side effect.
> 
> Two questions:
> 
> 1. Do you see any failures that this commit fix? If so, it's worth describing in the commit message. Regardless of the answer to this question, I think this PR change is nice.

Yes the following will be shown in the dmesg logs during cleanup.
```
[Wed Aug 19 21:20:25 2026] block nvme2n1: no usable path - requeuing I/O
[Wed Aug 19 21:20:25 2026] block nvme2n1: no usable path - requeuing I/O
...
```

On a side note `test 057` doesnt check dmesg logs or if the links are being used, thus making it always pass. test 057 also sets all the links to `inaccessible` causing the same dmesg logs as above. 
 
> 2. I found nvme/032 also kills the background _run_fio_rand_io(). Do you think it needs fix also? If so, would you mind creating another PR (or include in this PR)?

Ill create another PR. I'm not sure how i missed 32, but it might be because it only happens when `_nvme_disconnect_subsys` is called in the test as it starts cleaning up before `fio` finishes. any test that doesn't do a cleanup gets all of its subprocesses killed on exit and wont have this issue.

kawasaki:
@Mr-Bossman Thanks for covering the nvme/032 too!

PR #267, opened 2026-08-24
nvme: add support for nvme-cli 3.x commands

igaw:
nvme-cli 3.x grouped commands into plugins, such as all log
commands. The CLI supports the old commands via a trampoline command
implementation, but warns the user by writing an informational message
to stderr, which will cause many of the nvme tests to fail.

Add an indirection layer for nvme-cli commands that maps the new command
pattern to the old one, so support for nvme-cli 2.x can be dropped in
the future.

### DIFF
--- a/common/nvme
+++ b/common/nvme
@@ -38,58 +38,160 @@ _nvme_def_file_path() {
 	echo "${TMPDIR}/img"
 }
 
-_nvme_kxchap_cmd_style() {
-	if [[ -z "${_nvme_kxchap_style:-}" ]]; then
-		if nvme keys gen-kxchap --nqn nvmf-test-subsys > /dev/null 2>&1; then
-			_nvme_kxchap_style="3"
-		elif nvme gen-dhchap-key --nqn nvmf-test-subsys > /dev/null 2>&1; then
-			_nvme_kxchap_style="2"
+_nvme_cli_detect_version() {
+	if [[ -z "${_nvme_cli_version:-}" ]]; then
+		local ver major
+
+		ver="$(nvme version 2> /dev/null | awk '/^nvme version/ {print $3}')"
+		major="${ver%%.*}"
+		if [[ "${major}" =~ ^[0-9]+$ ]] && ((major >= 3)); then
+			_nvme_cli_version="3"
 		else
-			_nvme_kxchap_style="none"
+			_nvme_cli_version="2"
 		fi
 	fi
 }
 
-_nvme_tls_cmd_style() {
-	if [[ -z "${_nvme_tls_style:-}" ]]; then
-		if nvme keys gen-tls --subsysnqn nvmf-test-subsys > /dev/null 2>&1; then
-			_nvme_tls_style="3"
-		elif nvme gen-tls-key --subsysnqn nvmf-test-subsys > /dev/null 2>&1; then
-			_nvme_tls_style="2"
-		else
-			_nvme_tls_style="none"
-		fi
+declare -Ag _NVME_CMD_MAP=(
+	["keys gen-kxchap-secret"]="gen-dhchap-key"
+	["keys check-kxchap-secret"]="check-dhchap-key"
+	["keys gen-tls-psk"]="gen-tls-key"
+	["keys check-tls-psk"]="check-tls-key"
+
+	# log plugin
+	["log supported-pages"]="supported-log-pages"
+	["log telemetry"]="telemetry-log"
+	["log fw"]="fw-log"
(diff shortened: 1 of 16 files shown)


igaw:
FWIW, we decided to move the nvme-cli 3.0 release for a week for multiple reasons. The planned release is 2026-10-07 now

kawasaki:
@igaw Thanks for this PR. I did trial runs with nvme-cli versions 2.16 and 3.0, and both results were positive. Looks good from test run point of view. I will look into the code changes.

@guixinliu1995 FYI, the second commit of this PR touches nvme/054. The commit limits the reserve report filed checks only to three fields: gen, rtype and regctl regstrnt. This relaxes the test condition, but I think it is reasonable. If you have any comments, please share.

common/nvme
```diff
@@ -38,58 +38,160 @@ _nvme_def_file_path() {
...
+	if [[ -z "${_nvme_cli_version:-}" ]]; then
+		local ver major
+
+		ver="$(nvme version 2> /dev/null | awk '/^nvme version/ {print $3}')"
+		major="${ver%%.*}"
+		if [[ "${major}" =~ ^[0-9]+$ ]] && ((major >= 3)); then
+			_nvme_cli_version="3"
 		else
-			_nvme_kxchap_style="none"
+			_nvme_cli_version="2"
```
kawasaki, line 50: In theory, nvme-cli can have value 1, and in the future it can have value 4. Then, it is rather weird for me to set only 2 or 3 to the variable _nvme_cli_verion. How about to use the variable "_is_nvme_cli_pre_v3" instead? I think the code change below will do it.

```
diff --git a/common/nvme b/common/nvme
index b565d38..3827d8d 100644
--- a/common/nvme
+++ b/common/nvme
@@ -39,15 +39,15 @@ _nvme_def_file_path() {
 }

 _nvme_cli_detect_version() {
-       if [[ -z "${_nvme_cli_version:-}" ]]; then
+       if [[ -z "${_is_nvme_cli_pre_v3:-}" ]]; then
                local ver major

                ver="$(nvme version 2> /dev/null | awk '/^nvme version/ {print $3}')"
                major="${ver%%.*}"
                if [[ "${major}" =~ ^[0-9]+$ ]] && ((major >= 3)); then
-                       _nvme_cli_version="3"
+                       _is_nvme_cli_pre_v3=0
                else
-                       _nvme_cli_version="2"
+                       _is_nvme_cli_pre_v3=1
                fi
        fi
 }
@@ -149,7 +149,7 @@ declare -Ag _NVME_CMD_MAP=(

 _nvme_cmd() {
        _nvme_cli_detect_version
-       if [[ "${_nvme_cli_version}" == "3" ]]; then
+       if ((!_is_nvme_cli_pre_v3)); then
                echo "$1"
        else
                echo "${_NVME_CMD_MAP[$1]}"
@@ -170,7 +170,7 @@ _nvme_run() {
 _nvme_gen_kxchap_secret() {
        _nvme_cli_detect_version

-       if [[ "${_nvme_cli_version}" == "3" ]]; then
+       if ((!_is_nvme_cli_pre_v3)); then
                local -a kxchap_args=()

                while (($#)); do

kawasaki:
I made a couple of comments. Other than that, I think the code change look good to me. Thanks!

igaw:
> I made a couple of comments. Other than that, I think the code change look good to me. Thanks!

I don't see them. IIRC, depending how you comment you need to publish them explicitly in the github UI.

kawasaki (commented):
It would be the better to note in the commit message that this commit overrides the change by the commit 98b7f0360d7b ("nvme: add support for nvme-cli 3 tls/chap key
management"). Yeah, I think this commit is much cleaner :)

kawasaki:
> > I made a couple of comments. Other than that, I think the code change look good to me. Thanks!
> 
> I don't see them. IIRC, depending how you comment you need to publish them explicitly in the github UI.

Ah, I didn't push the "Submit Review" button. Now I have done it. Hope you can see them.

schlad:
For whatever it's worth I tested the PR against openSUSE Tumbleweed through our internal test framework; without the PR  with the latest `nvme-cli` we got some tests failing. The PR solved those tests issues and all the tests we run are passing well with it.

common/nvme
igaw: Good call! Makes indeed no sense to match on exact version number. I'll update the patch according your proposal.

kawasaki:
Thanks for the updates. I have merged this PR. Congrats for the nvme-cli v3.0 release :)

igaw:
Thanks a lot!

